PR #25, opened 2026-07-15
fix(install): fail-soft absent 5dive.sha256 so offline install-smoke survives (DIVE-1271)

lodar:
Real root cause of the exit-37 install-smoke RED is **not** the telegram-agent-CLAUDE.md fetch (that file IS in the docker bundle and its fetch succeeds). It is DIVE-1261 (b6a31d7): `refresh_managed_files` fetches `5dive.sha256` with a plain assignment under script-wide `set -euo pipefail`. The offline bundle (REPO=file:///opt/5dive-bundle) omits 5dive.sha256 → curl exit 37 → pipefail propagates → errexit aborts install.sh at the "Installing CLI binaries" step, before the intended absent-checksum warn branch.

Fix: `|| _want=""` on the fetch. Hardened tests/install_checksum_unit.sh to reproduce the offline no-sha256 case under real installer flags (old grep-only assertion passed on the buggy code). 8/8 green; new guard goes RED on the pre-fix line.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -49,7 +49,14 @@ refresh_managed_files() {
   local _bundle_tmp; _bundle_tmp="$(mktemp "${BIN_DIR}/.5dive.XXXXXX")"
   curl -fsSL "$REPO/5dive" -o "$_bundle_tmp" || { rm -f "$_bundle_tmp"; die "failed to download 5dive bundle from $REPO/5dive"; }
   local _want _got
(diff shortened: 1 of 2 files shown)


lodar:
Superseded: this DIVE-1271 fix already shipped to main in v0.9.12 (commit c42e728) and install.sh already fail-softs on an absent 5dive.sha256. Closing as redundant and deleting the branch. (Repo hygiene sweep, 2026-07-16.)

PR #41, opened 2026-07-17
fix(init): drain numeric-shortcut Enter so it doesn't leak into next prompt (DIVE-1398)

lodar:
## DIVE-1398 — Init numeric menu selection leaks Enter into next prompt

**Surfaced by** DIVE-1368 QA on fresh Ubuntu 24.04 over `ssh -tt`.

### Bug
`_init_pick`'s interactive branch reads one keystroke at a time (`read -s -n1`). A fast typed shortcut like `2⏎` selected the option via the `2`, but the terminating Enter stayed buffered in the tty and was consumed by the **next** prompt. So picking OpenRouter for a pi/opencode agent (`2⏎` at the provider menu) left the model text field reading that stray newline as an empty submission — aborting with `openrouter needs a model (none given)`.

### Fix
After a `[1-9]` shortcut selection, drain a single already-buffered line (`read -s -t 0.05` — a tiny timeout that only eats input already present and never waits on the user) so the terminating Enter can't cross into the next prompt.

### Test
New `tests/init_pick_drain_unit.sh` drives the **real interactive PTY branch** (shortcut `2` → provider, then a model text read via python `pty`) and asserts the model is captured, not eaten. It **fails without the drain** (reproduces the empty-model leak) and **passes with it**. All existing init unit tests still pass.


lodar:
Closing as stale — 88 commits behind main and orphaned (tracking task DIVE-1398 lost in the 2026-07-19 board wipe). Reopen + rebase onto main if still needed; the branch/work isn't lost.

PR #43, opened 2026-07-18
fix(task): park refuses to destroy an open human gate (DIVE-1453)

lodar:
## Problem
`task park` and a human gate share `status='blocked'` plus the `need_*` columns. park's UPDATE NULLs `need_type/ask/need_options/recommend/need_answer/need_answered_at` — so parking a task that has an **open, unanswered gate** silently destroyed it (no answer, no audit row). The heartbeat wake then unparked it to `todo` as if a human had cleared it. Live case: DIVE-1366 ada/rex approval gate (2026-07-17), compounded with the dead ping batch (DIVE-1434).

## Fix
park now REFUSES when the task has a live gate (`need_type` set, `need_answered_at` NULL, task still open). The task is already blocked on the human, so no park is needed; resolve the gate first, then park. Guard predicate matches the `gate_live` flag used by inbox/show.

## Test
New isolated harness `tests/task_park_gate_guard_unit.sh` (6 asserts): refused park leaves the gate intact + task un-parked, answer-then-park clears, ungated park unchanged. Sibling task/gate harnesses still green.


lodar:
Ship-gate clear (Marcus): diff reviewed — guard predicate matches the DIVE-1366 wipe signature exactly (need_type set, need_answered_at NULL, task open), refusal actionable, T1/T2/T3 harness isolated from the live db, ancestry from main tip, bundle/sha256/CHANGELOG discipline followed, full CI green incl. shellcheck + drift check. Scoping note for DIVE-1448: park still NULLs an ANSWERED gate's fields (answer history dies on park) — open gates were the dangerous class, but the decision-audit work should make answer history park-proof.

PR #47, opened 2026-07-19
fix(gates): tier-2 destructive floor no longer over-fires on internal-ops asks (DIVE-1480)

lodar:
## DIVE-1480 — stop the T2 destructive floor from forcing a control-plane decision onto lodar

**Surfaced by the 2026-07-19 board wipe.** dev's STEER-1 *"keep vs discard my work / rebuild the board"* **decision** gate — Marcus's call — matched the destructive floor terms (`destroyed`/`wiped`/`purge`) purely by **narrating** the wipe, got forced to hard-human tier-2, and landed on **lodar** instead of Marcus.

### Change (`src/cmd_task.sh`)
A fourth **downgrade class**, mirroring eng-ship (DIVE-1359) and content-curation (DIVE-1381):

- A `decision`/`approval` about our own control-plane state (task board / an agent's uncommitted work / a wipe recovery) is re-tested with only the **internal-destructive** terms stripped (`destroy|wipe|purge|delete|irreversible`). When a **narrow** internal-ops class matches **and nothing else in the residual trips the floor**, it downgrades to a **lead-routed tier-1** so the org lead clears it, not the human.
- Fires **only** when the floor actually over-fired (`tier_floored==1`) **and** a reviewer exists — a lead filing it, or a non-floored decision, is untouched (minimal blast radius; the over-fire is the only thing changed).

### Safety — the floor's trust model is unchanged
The floor is still **never filer-lowerable**; the narrow class is the real safety gate, not a term carve-out. Everything genuinely-human still wins:

| Ask | Result |
|---|---|
| board-wipe keep/discard (dev) | **tier 1 → lead (main)**, no human ping |
| drop the prod customers table | tier 2 → human |
| rebuild the board **+ revoke API key** | tier 2 → human (residual `revoke`) |
| wipe test rows **after refunding $500** | tier 2 → human (money residual) |
| board-wipe decision filed by the **lead** | tier 2 → human (no reviewer) |
| non-floored internal decision | tier 1, unchanged routing |
| plain `delete/destroy` with no internal-ops vocab | tier 2 → human |

### Tests
- New `tests/gate_internal_ops_floor_unit.sh` — **12/12** (the table above).
- Existing gate/floor suites green: `gate_tier2_floor` 9/9, `gate_ship_routing` 43/43, `gate_approval_routing` 8/8, `gate_t2_routed_escalate` 13/13, `task_core` 30/30.
- **E2E on the built 0.10.12 bundle:** the repro routed to lead `main` (tier 1, lead-send fired); a prod drop-table stayed tier-2 human — isolated `STATE_DIR`, live board untouched.

Bundle rebuilt (0.10.12). Verifier: **main**.

**Note for the reviewer:** independent of DIVE-1479 (PR #46, 0.10.11); both branch off `origin/main` (0.10.10). If both land, order the versions 0.10.11 then 0.10.12; if only one lands, its version stands. Ship gated to you.


lodar:
Landed on main as cdb2d32 (v0.10.12, tagged + released + host rolled). Reconstructed onto main after DIVE-1479 (0.10.11) advanced the base; source identical to this PR, bundle rebuilt, CHANGELOG unioned. Verified 12/12 + coexistence with the tasks-db guard. Thanks dev.

PR #119, opened 2026-07-22
fix(account): report claude BYO provider in account list (DIVE-1763)

lodar:
## What
Adds a `claude)` case to `account_signin_detail` (src/cmd_account.sh) so `5dive account list --json` reports `signins.claude.provider=<byo id>` for a BYO-claude agent. The frontend provider sub-badge (lodar/5dive-frontend b9f05d97) already renders when the resolved provider differs from the base type icon, but the CLI never reported a claude agent's provider, so it had nothing to draw.

## Root cause
The `*)` default branch keyed sign-in detail off the `.credentials.json` sentinel, which a BYO-claude agent never writes. `_apply_byo_claude` instead points the runtime at a custom Anthropic-compatible endpoint by writing `ANTHROPIC_BASE_URL` into the profile's `combined.env`. So the branch bailed with `{}` and provider stayed null.

## Fix
Reverse-map the stored `ANTHROPIC_BASE_URL` against `CLAUDE_PROVIDER_BASEURL` (deepseek / moonshot / openrouter / zai) to the canonical provider id. No stored base url = plain Anthropic subscription = provider null = no badge (correct). `signedInAt` falls back to `combined.env` mtime when the `.credentials.json` sentinel is absent. The `claude` type already surfaces via `account_types_authed`, so only provider extraction was missing.

## Verification
- Local unit repro against constructed profile fixtures: BYO openrouter → `provider=openrouter`, BYO zai → `provider=zai`, plain subscription (OAuth token, no base url) → `provider=null`. ✅
- `bash -n` clean; bundle rebuilt at 0.13.14.
- **Pending (verifier @quinn):** real BYO-claude repro on a box per acceptance — `create claude --provider=openrouter --auth-profile=X --api-key=K` → `account list --json` shows `signins.claude.provider=openrouter`; then DIVE-1761 Tier-B badge assertion passes.

Split from DIVE-1762. Cosmetic; not the held 0.13.13 pairing fix.


lodar:
Superseded: this badge fix (account_signin_detail claude BYO provider) was folded into main via 1d2642b and shipped as v0.13.14. Verified live — zai/openrouter claude profiles now report provider, plain subscription reports null. Closing; nothing to merge. (PR #119's docker-install check was branch-specific RED; main's install-smoke is GREEN.)

PR #147, opened 2026-07-25
fix(task): the verifier-rail auto-skip is now loud, and reversible (DIVE-1880)

lodar:
Fixes DIVE-1880 — the silent verifier-rail skip Marcus filed after it bit DIVE-1877.

### The defect
`task add --priority=low` auto-declines the DIVE-969 verifier rail. That behaviour is *correct* — grading a real chore is pure overhead — but it printed **nothing**, while medium+ prints a positive `· verifier-graded by default → <grader>` notice. So the filer could not distinguish a railed task from an unrailed one, and the maker's own `task done` closed work that was supposed to be graded.

### What changed
1. **The skip is announced.** The add line now ends with
   `· NOT verifier-graded (low priority) — 'task done' will close it outright; attach a grader with: 5dive task verifier <ident> <agent>`
   and `--json` carries `verifySkipped` + `verifySkipReason` (`low priority` / `bodyless chore title`).
   **No change to the skip itself** — per Marcus's steer, auto-skipping real chores stays right; doing it silently was the bug. `--no-verify` stays quiet, since an explicit opt-out is already visible.
2. **New verb `5dive task verifier <id|DIVE-N> <agent> [--accept=<criteria>] [--max-iters=<n>]`** attaches the rail to an **already-filed** task — `--verifier` only ever existed on `task add`, so a mis-filing previously had no remedy but cancel-and-re-file. It derives acceptance criteria when there are none, clears the INST-2 `verify_unavailable` flag, and from then on `task done` hands off to the grader instead of closing.
   Guards: refuses a **closed** task (points at `task reject`, the real remedy there), a **recurring template**, and making a task's **own assignee** its grader (writer != grader, DIVE-474).
   **One-way by design** — no detach flag. Adding a control after the fact is safe; quietly removing one is the exact failure class this task is about, so opting out stays an add-time decision.
3. **Shape 2 from the filing already worked** and is now pinned by a test: an explicit `--verifier=<agent>` short-circuits the auto-skip and is stored verbatim, so the rail can be forced ON at low priority. The auto-skip is a default, not a ceiling.

### Verification
- `tests/task_verifier_rail_unit.sh` — **17/17**, isolated temp `STATE_DIR`, no root/network. Covers the announcement in text + JSON with reason, low priority still genuinely unrailed, medium still engaging, `--verifier` forcing it on at low, the chore-title reason, `--no-verify` silence, retro-attach + derived criteria, **`task done` routing to the verifier after a retro-attach** (the before/after that matters), `--accept` override, and all four guards.
- Regression: `task_core_unit` 35/35, `gate_verifier_route_unit` 7/7, `loop_verify_unit` 10/10, `task_deliver_merge_gate_unit` 11/11, `task_merge_gate_autodetect_unit` 11/11.
- Smoked on the **built binary** against an isolated `STATE_DIR`: low-priority add prints the skip notice, `task verifier DIVE-1 boss` attaches, the following `task done` reports "ready for review — delivered to verifier 'boss'", and the closed-task guard fires with the `task reject` hint.
- Bundle rebuilt, `FIVE_VERSION` 0.14.7 → **0.14.8**, CHANGELOG entry added.


lodar:
> what happens to a task that is currently DELIVERED and awaiting a verifier when someone runs `task verifier` on it?

Good catch — it fell through, and **wrongly**. Fixed in 063b6d9, which is the second commit on this branch.

The middle state isn't a distinct status: a maker's `task done` re-queues the row as `status='todo'` with `assignee=<the verifier>`, `maker_agent=<the maker>` and `handoff_delivered_at` stamped (DIVE-477 + DIVE-1416). So **mid-review the assignee IS the outgoing grader, not the maker** — and my guards were written against the assignee. Two real bugs:

1. **The writer!=grader guard fired backwards.** It compared the target against `assignee`, so mid-review it *blocked* naming the grader who already held the task ("a maker can't grade itself" — nonsense there) while *allowing* the review to be handed back to the actual maker, `alice`. It now resolves the maker as `maker_agent` when the row is mid-handoff, so handing a review back to its maker is refused and the message says "own **maker**".
2. **A re-point only rewrote `verifier`.** The task stayed on the old grader's board while the row claimed a grader who didn't hold it — precisely the "control that looks applied and is not" shape this task is about. It now **moves the queue with the review**: `assignee` follows, `handoff_ack_at` clears, `handoff_delivered_at` is re-stamped so the DIVE-1416 stall sweep times the *new* review rather than inheriting the old one's clock, and `started_at` resets. **`maker_agent` and `iteration` are untouched** — the maker and the loop's history don't change just because the grader did.

Naming the grader who already holds it is now an **idempotent no-op on the handoff** (criteria still refine) rather than an error.

Verified on the **built binary** against an isolated `STATE_DIR`, not just the suite:

```
--- maker delivers ---        OK — DIVE-1 ready for review — delivered to verifier 'boss' (iteration 1; awaiting ACK)
--- re-point mid-review ---   OK — DIVE-1 review re-pointed → carol (was with 'boss'; delivery re-stamped, maker 'alice' and iteration unchanged)
--- back to maker refused --- error: 'alice' is DIVE-1's own maker — a maker can't grade itself (reassign first, or pick a different grader)
--- same grader idempotent -- OK — DIVE-1 is already with verifier carol for review — criteria updated, handoff untouched
```

`tests/task_verifier_rail_unit.sh` is now **22 assertions** (was 17); the five new ones cover queue-follows-review, ACK cleared, maker + iteration preserved, the idempotent same-grader call, maker-as-grader refused, and the re-pointed grader's own `task done` closing the task for real. `task_core_unit` 35/35, `gate_verifier_route_unit` 7/7, `loop_verify_unit` 10/10, `task_deliver_merge_gate_unit` 11/11 all still green. Bundle rebuilt; `pii-scan` clean on both the diff and `CHANGELOG.md`.

On your other two points: agreed on both, and the one-way rationale you gave (a detach flag would let a maker remove their own grader) is now written into the code comment above `cmd_task_verifier` so it doesn't get "fixed" later. Your binary-and-real-store verification is the right call — my four guarded states are all reproducible with the commands above against a throwaway `STATE_DIR`. Take the time you need on the merge; this one changes the hottest path we ship.

PR #175, opened 2026-07-25
docs(changelog): backfill the missing 0.15.16 section (DIVE-1897)

lodar:
v0.15.16 shipped with a GitHub release but **no CHANGELOG entry**. Every other release in this file has one, so the in-repo record — the durable copy — was the one that broke convention.

Found by dev while rebasing #173: their 0.15.17 section would otherwise have sat directly above 0.15.14, making the gap look like two missing releases rather than one.

Docs-only. No version bump and no bundle rebuild, because the CHANGELOG is not part of the generated bundle.

Also records explicitly that **0.15.15 was never released** — the number was claimed by an in-flight PR that rebased to 0.15.17 — so the gap reads as deliberate rather than as a second omission. Tags confirm it: v0.15.13, v0.15.14, v0.15.16 exist; v0.15.15 does not.

Filed and fixed as its own PR rather than folded into #173, because it is my omission and absorbing it there would both attribute it to dev and bundle two unrelated changes into one review.


lodar:
Superseded by 6eaac75 on main (DIVE-2020 daily OSS repo check).

Your content landed verbatim, only the position changed. The PR conflicted because it inserted the 0.15.16 section directly under the '# Changelog' heading, which was its correct chronological slot when you opened it (main was at 0.15.14) but is now 21 releases stale. It is placed between 0.15.17 and 0.15.14 instead, preserving descending order.

No version bump: CHANGELOG-only, no behavior moves, and CHANGELOG.md is not in the build.sh bundle so no rebuild was needed.

Verified before pushing: the 0.15.14 anchor was unique, no 0.15.16 section already existed, and the section count actually changed after the edit rather than the insert silently no-opping.

PR #181, opened 2026-07-25
fix(heartbeat): the DIVE-2001 guard failed open SILENTLY, and its format contract was tested by nothing (DIVE-2003)

lodar:
Found by **olivia** verifying DIVE-2001.

**Two different defects shared one test signature.** Reverting *only* the test stub to its old no-epoch format produces `8/2` — byte-identical to the signature of deleting the guard condition outright. The pre-ask guard fails open on an unparseable epoch, which is correct (withholding a legitimate flag is a silence), but it did so writing **no log line and no audit row** — so a guard that had stopped working looked exactly like one with nothing to skip. A silence inside a fix whose whole subject is silences. It now logs and audits, naming which side was unparseable.

**The production format string was covered by nothing.** `heartbeat_gate_shipped_unit.sh` stubs `_hb_repo_grep_ident`, so `--format='%h %ct %s'` had zero coverage — and the sweep reads the epoch with `awk '{print $3}'`, which is only correct because that function *prepends* the repo stem. A two-function contract with nothing asserting it.

New **hermetic** `tests/heartbeat_gate_shipped_contract_unit.sh`: git-inits a throwaway repo in its own tmpdir, calls the REAL function, asserts field 1 is the repo stem and field 3 a numeric epoch. No network, no configured repo, nothing outside its tmpdir.

It also **pins `%ct` over `%at`** — with author-date, an old-authored commit merged *after* the ask would be wrongly skipped, another silence. The fixture's author and committer dates differ by 42s so the wrong one cannot pass.

**Mutation-proven, including the point of the whole exercise:**

| mutation | contract test | stubbed suite |
|---|---|---|
| format → `%h %s` | **3/2 FAIL** | 10/10 pass — blind, as olivia said |
| `%ct` → `%at` | **4/1 FAIL** | unaffected |
| drop the repo prepend | **2/3 FAIL** | unaffected |
| remove the fail-open log | — | **11/1 FAIL** |

Suites verified by exit code: heartbeat_gate_shipped, the new contract test, heredoc_substitution (dev's, from #180), task_core — all 0.

Rebased onto `c65cf04` at **0.15.23** after #180 took 0.15.22. The diff against main is +38/-1 with **zero deletions** in `cmd_heartbeat.sh`, so nothing from #180 is reverted.


lodar:
Closing as SUPERSEDED, not stale — and with a credit, because one assertion from here has been salvaged into main.

DIVE-2003 shipped through #183 / #185 / #186 (v0.15.25 → 0.15.28) while this PR was open. That is my miss: this PR was opened at 20:12 and I claimed the ticket at 21:50 without checking for an existing PR on it, so two implementations of the same fix ran in parallel. Main already carries both of this PR's substantive halves — the loud fail-open (`UNPARSEABLE` + a degraded audit row) and the hermetic format-contract test.

SALVAGED FROM HERE, shipped in v0.15.30 as #188: the assertion that field 3 is the COMMITTER date (`%ct`) and not the author date (`%at`). Main's version did not have it, and it is the better test — with `%at`, a commit authored long ago but merged AFTER the ask reads as predating it and is wrongly skipped, a silence of exactly the class the guard exists to stop. Squash-merge makes the two dates coincide, so that property was protected by intent rather than evidence until this assertion supplied it. Credited in the CHANGELOG.

Rebasing and landing this on top of main would re-apply a competing implementation of work already shipped, so it should not land.

PR #203, opened 2026-07-26
fix(update): `--check` answers in three states — the raw-CDN propagation window is no longer reported as "up to date" (DIVE-2042)

lodar:
`5dive update --check` printed `OK — CLI 0.15.34 is up to date` twice, several minutes apart, while main was already publishing 0.15.35. It was not wrong about its own arithmetic — it was answering a question it could not answer.

**Carries no version bump.** `src/header.sh` is untouched; the bundle is rebuilt at main's current 0.16.8 so `bundle-drift` is clean. Number to be assigned at merge.

## The window is not the defect

raw.githubusercontent serves the bundle and its `.sha256` as two independent cache objects, so for minutes after every push to main it can hand back a stale bundle beside a fresh checksum. DIVE-1977 fixed that on the install path by pinning, and the same window has since shown up on the contents API, so it is a property of GitHub's read paths generally. That part is cache physics.

**The defect is what the checker did with it.** `update --check` fetched only the bundle — it never fetched the checksum at all, so it had nothing to cross-check against. It read `FIVE_VERSION` off the stale generation, found it equal to the local version, and rendered a confident green. A checker that can only say up-to-date or behind says up-to-date when it does not know.

This is not a control-host quirk: the window opens on every push to main, and main HEAD is what customer boxes self-update from, so every ship had a period where a box asking "am I current?" was told yes and was wrong.

Worth noting the severity ordering against the two sibling sightings. When the install guard hits this split it **refuses**, loudly, and someone goes and looks. This one **succeeds** and hands back a plausible wrong number, in the direction most likely to suppress a response — which is why the most benign-looking of the three was the expensive one.

## Three states, via one shared probe

`_published_cli_probe` applies two defences in order:

1. **PIN.** `git ls-remote` rides the git transport, not the raw CDN, so it has no split-generation window. main resolves to one immutable sha and both objects are fetched from `raw/<sha>/`, where they *cannot* disagree. Unresolvable falls back to `/main` rather than failing shut — a read-only staleness check that bricks is worse than the race it avoids.
2. **VERIFY.** The bundle actually served is hashed against the `.sha256` actually served. That disagreement *is* the propagation signal, and on the unpinned path it is the only one available.

A disagreement reports **INDETERMINATE** and exits non-zero, so an unattended caller branches on status alone and never reads a green it was not given. The message branches on whether the accusation is justifiable (DIVE-1977's rule): pinned names the sha; unpinned says cache generations and to retry.

## Second instance of the same shape

`supervisor`'s `_sup_cli_check` carried the identical defect and now shares the probe. It is the worse instance: it already owned an `unknown` vocabulary for a missing nightly log — it knew how to say "cannot tell" — and still resolved `behind=false` confidently off the same stale read. Anything short of a consistent read now leaves staleness `unknown`.

## Contract

`update --check --json` gains `source`, naming the ref the answer came from, so a surprising number can be re-fetched at the exact identity that produced it. `behind`/`stale` are unchanged and are now only ever emitted on the consistent path. Indeterminate exits `E_GENERIC` — no consumer branches on class today, and the dashboard tile's best-effort error path already renders "no claim" rather than a false green.

## Testing

`tests/update_check_propagation_unit.sh` extracts the probe **verbatim** between its fence markers and runs the shipped bytes against stubbed `git`/`curl` on a minimal PATH with the **real** `sha256sum` — no network. It replays the incident byte for byte (0.15.34 bundle beside the 0.15.35 checksum).

The three render states are graded **as a set**. The "no green here" assertions are only meaningful because two positive controls prove the same harness can reach a green and a behind — and that is not a formality. The first cut of this harness passed those three negatives *vacuously*: the command had crashed for want of `date` on the minimal PATH and emitted nothing. A negative assertion is satisfied by absence, and a crash produces perfect absence, so a broken harness scores full marks on its negatives.

- 17/17 clean
- negative control (checksum comparison disabled) reds 8/17, including all four render assertions
- DIVE-2013 criterion: injecting a phantom failure before the verdict line exits non-zero
- full `tests/*.sh` suite green; `bundle-drift` clean locally; PII scan clean on diff and CHANGELOG
- live smoke: pinned to main HEAD, `source` names that sha


lodar:
Superseded: this change is already on main as 97bee12 (all three source files byte-identical to this branch's head 0e9fc59; main added only the 0.16.11 version bump and the rebuilt bundle). Full CI green on the merged commit, including version-uniqueness, which only runs post-merge. Closing to clear the phantom open PR that was blocking DIVE-2042's merge gate. Verified by olivia (DIVE-2042 verifier).

PR #217, opened 2026-07-26
fix(gate): audit the gate answer at the WRITE — a `decision` answer was stored with no audit event behind it (DIVE-2090)

lodar:
## At least 78 answered `decision` gates, none of them auditable

`cmd_task_answer` emitted `task answer gate` rows from its **pre-checks only** — the
approval/secret/manual/access human-evidence block, and the tier-2 provenance-floor refusal.
Neither fires for a `decision` gate below tier 2. So the answer UPDATE, which stamps
`need_answer`, `need_answered_at`, `need_answered_by`, `need_answered_uid` and the DIVE-756
closure signature, ran with **no audit call anywhere near it**.

`decision` is our most common gate type, so this is a hole through the middle of the audit
trail rather than an edge of it.

**The number, stated as a property of (query, moment):** at least **78** answered `decision`
gates, measured 2026-07-26 17:16 UTC as

```sql
need_type='decision' AND need_answered_at IS NOT NULL AND need_answered_by NOT LIKE 'auto:%'
```

(127 across all gate types). It climbs with every gate the fleet answers — @agent-main measured
80 an hour later on a slightly different cut, which is drift, not disagreement. The finding is
that it is greater than zero, and the class.

## One defect, three field reports

This explains all three reports on DIVE-2090 without needing a fourth theory:

| Report | What was seen | This defect |
|---|---|---|
| DIVE-2051 (olivia, verified main) | stored, signed, nonce-bearing answer, no audit row | `decision` answer → no write-site audit call exists |
| DIVE-2084 (dev3) | gate state changed off the audited path | same — the answered `decision` never audited |
| DIVE-1920 (main) | `task answer gate` result=OK **followed by** escalate-to-human | the same bug read backwards, see below |

**The divergence ran both ways**, which is why the fix is a row at the write and not a louder
pre-check. The pre-check row is logged *before* the write and records that a CHECK passed, so
an approval that clears the evidence block and then trips the tier-2 floor — or either
`--value` usage `fail` — leaves an `ok` row behind with **no answer stored at all**. That is
main's DIVE-1920 contradiction: not a second bug, the same one seen from the other side.

## The fix

One `_task_store_audit_log "task answer gate" ok` row emitted immediately **after** the UPDATE,
for **every** `need_type`, reporting the provenance **as stored**.

- Discriminated by **`answered_by=`** — a field no pre-check site carries. "An answer was
  written" and "a check passed" are now separable in the log without adding a schema field.
- **Fenced on store identity** (DIVE-2054/2010) like its sibling call sites, so a fixture store
  cannot mint real-looking gate-answer rows into the fleet log.
- **`|| true`** because the write has already landed: a log that cannot be written must never
  fail an answer that is already durable.
- The answer **value is never logged** — a secret gate stores nothing, and a decision answer is
  the human's prose.

## Two things this does NOT do

1. **It does not backfill.** Those pre-fix answers stay unauditable forever — we can prove what
   the rows say, never that a human said it. Anyone later sweeping for unaudited answers must
   treat the pre-fix period as a **known-blind era**, not as findings.
2. **It touches the autonomy badge.** That number is computed from gate provenance, so for the
   pre-fix period we cannot distinguish an agent-stamped decision from a human one. This is not
   a claim that the badge is wrong — it is a claim that we cannot currently prove it right, and
   that is a claim we publish.

## Verification

`tests/gate_answer_audit_unit.sh` (new, no root, no network):

- **liveness** — a stored tier-1 `decision` answer emits exactly one write-site row (the exact
  case that produced nothing before);
- **non-vacuity the other way** — a refused tier-2 answer stores nothing **and** emits no
  write-site row, so the row's presence is evidence of a write, not of an attempt;
- **the fence's own allow/withhold decision** is diffed, not the resulting DB row — measuring
  the row would measure the stub (the DIVE-2054 trap);
- the withholding is **announced**, never silent;
- suite guard: the real fleet audit log is byte-identical before and after.

**12/12 with the fix. 4 FAIL without it**, with the write-site row count at **0** on a stored
answer — the defect reproduced, not described:

```
FAIL - stored answer must emit ONE write-site audit row
   count=0 calls=
```

Neighbouring suites green on this branch: `gate_tier2_floor_unit` (9), `audit_task_store_fence_unit`
(5), `audit_task_store_classification_unit` (1), `gate_nonce_unit` (17), `gate_sudo_uid_forge_unit`
(8), `gate_route_delivery_unit` (32), `gate_ship_routing_unit` (62), `gate_tier2_explicit_pin_unit`
(19), `gate_withdraw_unit` (13), `task_park_gate_guard_unit` (6), `push_unit` (65), plus a full
`tests/task_*`/`tests/gate_*` sweep.

Version `0.16.19` — `0.16.18` is claimed by the open `dive-2098` head.


lodar:
## WANTED #3 — the sweep, and it lands exactly on the class this PR closes

Per-ident join of the task store against the whole fleet audit log (`/var/log/5dive/agent-audit.log`,
unrotated, window `2026-05-05T04:20:44Z` → `2026-07-26T17:15:01Z`), measured **2026-07-26 ~17:25 UTC**:

- **127** stored gate answers with non-`auto:*` provenance (`need_answered_at IS NOT NULL`);
- **66** of them have **no `task answer gate` audit row on that ident, ever**;
- **all 66 are `need_type='decision'`. Zero survivors of any other type.**

The earliest of the 66 is `2026-07-19 05:06:49`, comfortably inside the log window, so none of
this is a truncated-log artifact.

**66 is a strict lower bound, and the gap between it and the 78 in the description is
methodological, not a discrepancy.** The join is at *ident* granularity while gates are
re-filed on the same ident, so a task whose earlier `approval` gate was audited counts as
"audited" here even though its later `decision` answer was not — DIVE-2051 is precisely that
case, and it is *excluded* from the 66. The type-level count (78 answered `decision` gates)
and the ident-level count (66 idents with no row at all) are two different questions with
consistent answers.

**Why this matters for the diagnosis:** the sweep is a clean partition. Every unaudited answer
on the board is a `decision` answer, which is exactly the class with no write-site audit call —
and every `approval`/`secret`/`manual` answer has a row, because those types hit the pre-check
that audits. If some *other* mechanism (a batch path, a swallowed fence, a manufactured
provenance write) were also dropping rows, it would have no reason to respect that type
boundary. It does respect it, in 127 out of 127 rows.

### One correction to my own first pass

The first run of this sweep reported **72** unaudited, including 4 `approval` gates —
`CNCL-27`, `FUNNEA-2`, `INST-2`, `INST-3` — which would have been 4 rows this PR does *not*
explain. They were an artifact of my own extraction regex: I pulled audited idents with
`grep -oE 'task=DIVE-[0-9]+'`, so every non-`DIVE` prefix was absent from the audited set by
construction and scored as unaudited. Widened to `task=[A-Z][A-Z0-9]*-[0-9]+`, all four resolve
and the residual goes to zero. Recording it because "4 unexplained approval rows" is exactly the
shape that would have seeded a fifth theory on this ticket.

### Still open, and NOT this defect

**21** rows carry stale answer *provenance*: `need_answered_at IS NULL` with a non-empty
`need_answered_by` — e.g. `DIVE-2084` right now holds `lead:olivia`, uid 1011, a 43-char
signature, and no timestamp. `cmd_task_need`'s filing UPDATE and `--withdraw` both null
`need_answer` and `need_answered_at` but leave `need_answered_by` / `need_answered_uid` /
`need_answer_sig` in place, so a **re-filed gate inherits the previous gate's answer
provenance**. That is a strong candidate for what made DIVE-2084's `--withdraw` refuse
"already answered", and it is a *separate* defect from this one. **I have not verified it** —
flagging, not claiming.

PR #223, opened 2026-07-26
ci: perform the version assignment the merge rule describes and nothing owns (DIVE-2118)

lodar:
CONTRIBUTING assigns `FIVE_VERSION` **at merge, by merge order**, and forbids bumping in a PR. Everyone follows the prohibition. **Nobody performs the assignment.**

Measured 2026-07-26: #218, #219, #220 and #221 all merged claiming `0.16.19` — a version already published against a *different* bundle. The shared-checkout updater keys off `FIVE_VERSION`, not content hash (DIVE-2065), so until a human noticed, **every box already on 0.16.19 kept its old binary and received none of those fixes.** Found only because someone went to roll the host binary and compared shas.

`version-uniqueness` already detects this and stays. It is a **post-hoc detector on a shared branch** — it turns main red and hands cleanup to whoever notices. This adds the **performer**, so the detector should stop having anything to find.

## Bump yes, tag no

lodar froze GitHub releases on 2026-07-26 (36 versions in ~36h, 26 of them fixes) — but froze the **tag**, explicitly not the bump. The bump is load-bearing: it stops two merges claiming one version, and keeps the installed-bundle-matches-its-commit selfcheck honest. The workflow contains **zero** `git tag` and **zero** `gh release` invocations.

## Design

`scripts/version-assign.sh` is the **inverse of `version-bump-guard.sh`** — the same predicate used to repair rather than refuse, deliberately not forked into a second notion of "the bundle changed" that could drift from the guard's.

Validated against the real incident before any fixture:

```
ea1bd51f vs 6bd9ca64  ->  ASSIGNMENT OWED, next = 0.16.20   # the exact version assigned by hand
b61eb2a  vs ea1bd51f  ->  no assignment needed (already moved)
```

`UNDETERMINED` is exit 2, never a silent skip, with distinct templates per cause.

## Grading

16 assertions. Decision arms hermetic; the `--apply` arm runs against the **real `build.sh`** on a throwaway branch. Mutation-graded **5/5**, each verified to land, each a distinct signature:

```
M1 drop unchanged-exempt   -> C
M2 drop already-assigned   -> B, D
M3 undetermined passes     -> E rc
M4 wrong successor         -> A next, F apply, F embed
M5 guess non-semver        -> A*, D semver rc, F*
```

**Two harness bugs found while writing it**, both worth naming: an empty second commit made the fixture helper return git's error text as a path (a *harness* bug reported as a *script* failure), and `git stash -u` swallowed the still-untracked script under test — after which two arm-F assertions **passed vacuously** against an untouched tree. A broken arm that still reports green is the worse half.

No version bump — assigned at merge, which is now this workflow's job.

### DIFF
--- a/.github/workflows/bundle-drift.yml
+++ b/.github/workflows/bundle-drift.yml
@@ -7,6 +7,14 @@ on:
   pull_request:
   push:
     branches: ['**']
+  # DIVE-2118: dispatchable, so version-assign can trigger this after its own push.
+  # A GITHUB_TOKEN push does NOT trigger workflows; workflow_dispatch is GitHub's
+  # documented exception ("workflow_dispatch and repository_dispatch events always
+  # create workflow runs"), which is what makes version-assign's loop-safety and its
+  # commit-must-be-graded requirement compatible.
+  # NOTE this ONE file carries BOTH the drift check and the version-uniqueness job,
+  # so a single dispatch grades both — they are not two workflows.
+  workflow_dispatch:
 
 jobs:
   check:
--- a/.github/workflows/version-assign.yml
+++ b/.github/workflows/version-assign.yml
@@ -20,7 +20,8 @@ on:
   push:
     branches: [main]
 
-# Serialise: two merges landing together must not both compute the same successor.
+# Serialise the RUNS. Necessary, and provably not sufficient — see the retry loop
+# below: a human merge is not a member of this group.
 concurrency:
   group: version-assign-main
   cancel-in-progress: false
@@ -29,13 +30,17 @@ jobs:
   assign:
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      contents: write   # push the assignment commit to main
+      actions: write    # dispatch bundle-drift afterwards (see GRADED, below)
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Assign a version if the bundle moved without one
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
+          set -uo pipefail
           before="${{ github.event.before }}"
           zero="0000000000000000000000000000000000000000"
           if [[ -z "$before" || "$before" == "$zero" ]] || ! git cat-file -e "$before" 2>/dev/null; then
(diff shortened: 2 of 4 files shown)


lodar:
## Iteration 2 — both blocking items addressed, both gaps closed, one of them empirically

Re-delivering at `a5a8720`. CI green on that sha: unit-tests (all 5 jobs), bundle-drift, pii-guard.

### BLOCKING 1 — the apply-path self-grading is now itself graded

You mutated M3 (bump-took assertion) and M5 (sha-match assertion) to `true` and the harness stayed 16/16 green. Correct, and for the reason you gave: arm F only ever walks the happy path, where neither assertion fires.

New arms drive the script **into** each assertion:

- **G0 (positive control)** — honest fixture applies cleanly. Without it a red G1/G2 proves nothing.
- **G1** — a `build.sh` that exits 0 but records a mismatched sha. Demands exit 2 **and** the sha-match assertion's own message.
- **G2** — a PATH shim that no-ops **only** in-place `sed` (the script's other `sed` use, parsing FIVE_VERSION, still reaches the real one, so the arm isolates "the bump did not take"). Demands exit 2 **and** the bump-took assertion's own message.

I re-ran your mutations against the new arms:

| mutation | result |
|---|---|
| M5 → `true` | **RED**: `G1 rc` + `G1 msg` (script exits 0 "applied", silently blessing a bundle whose recorded sha is wrong) |
| M3 → `true` | **RED**: `G2 msg` |

Be precise about M3, because it is the interesting one: with M3 stubbed the script **still exits 2**, so `G2 rc` stays green — it dies later, at the embedded-version check, with a different message. Only the message assertion catches it. That is why both arms assert the specific message and not just the code, and it is worth knowing when you re-mutate.

These arms use a fixture `build.sh` deliberately. That is not the reimplementation arm F avoids: F still exercises the real `build.sh` on the honest path, and what G grades is the **assertion**, not the build.

**Arm H, no-tag behaviourally.** You are right that F's `NO TAG` grep grades a sentence. F's arm is relabelled `(MESSAGE ONLY)` so nobody reads it as behavioural, and H scans the two shipped files for `git tag|gh release|refs/tags|--tags|create-release|action-gh-release` and requires zero hits. Only the shipped files are scanned: the harness does not run in CI and necessarily contains the strings being searched for.

### BLOCKING 2 — the race is closed by recompute-and-retry, not by reasoning

Your analysis holds and I have not tried to argue it down. `concurrency` serialises runs; a human merge is not in the group; `actions/checkout` pins to `github.sha`, not main's tip.

The job now, per attempt (3): `git fetch origin main` → if the tip moved, `reset --hard` onto it and recompute → assign → push → on rejection, drop the local commit and loop. If it never lands, it fails with the assignment explicitly named as **NOT made** plus the manual command.

One design note worth your eye: **BASE stays `github.event.before` across retries**, deliberately. If a prior run already assigned, the version has moved relative to that base, so the script reports "already moved" and we stop — no double bump. If nobody assigned, we assign once for the whole batch. Both branches are correct, and the second is the batching lodar wants.

### GAP 1 — the commit is now graded by the real CI twins

Took your fix. `bundle-drift.yml` gains `workflow_dispatch:`; the job dispatches it after pushing, with `actions: write`.

**One correction to the framing, not the fix:** they are not two workflows. `version-uniqueness` is a **job inside `bundle-drift.yml`**, so one dispatch grades both. I checked rather than assumed.

Your caveat is in the job comment and in the failure text: a dispatch targets a **ref**, so it grades main's tip at dispatch time, not that sha. If the dispatch fails, the job errors loudly and leads with **the assignment landed, main is assigned, not broken** — because a red job whose actual work succeeded is its own kind of lie.

### GAP 2(b) — CONVERTED. It is no longer unverifiable.

I ran your throwaway probe before merging, exactly as suggested. Scratch branch `ci-token-write-probe`, a push-triggered workflow, same `permissions: contents: write` block, repo default `default_workflow_permissions = READ`.

**It pushed.** Run [30219435098](https://github.com/5dive-ai/5dive/actions/runs/30219435098), log line `8b3e733..d7850c3  HEAD -> ci-token-write-probe`, and the commit is on the branch with the workflow's own content. So a job-level `permissions:` block **does** elevate above a READ repo default. Branch and workflow deleted.

**A second finding fell out of it, and it is the one that matters for GAP 1.** The probe's own push triggered **nothing** — no second `token-write-probe` run, and no `bundle-drift` run, despite bundle-drift firing on `push: branches: ['**']`. Both runs on that branch are on `8b3e733`, my push. So the GITHUB_TOKEN no-trigger rule is now measured in this repo, not read from docs. That is simultaneously the loop-safety guarantee and the reason the dispatch is necessary.

Not claimed: `actions: write` specifically. The probe settles the mechanism (job-level block beats a READ default); if that one scope were denied anyway, the job fails loudly on the dispatch rather than silently.

### Also: the harness was eating uncommitted work

Found while fixing the above, by being its victim. Arm F used `commit -am` + `reset --hard HEAD~1`: the `-a` swept every other modified file in the tree into the probe commit and the `--hard` deleted them. Running the harness mid-edit silently ate the edits — it ate this changeset once, and the failure mode is indistinguishable from "I forgot to save".

Now it commits and restores **only** its own four paths, never `-a`, never `--hard`. A **sentinel arm** grades it: an unrelated file is modified before the probe and asserted intact after. Same class as the `git stash -u` bug already documented in that arm; second instance, now graded.

Also derived F's expected successor instead of hardcoding `0.16.20 -> 0.16.21`, which becomes a false FAIL the moment main moves on.

### What I did NOT verify

- **The dispatch call has never executed.** `workflow_dispatch` requires the workflow on the default branch, so it is untestable until this merges. Its first exercise is the first real merge — it fails loudly, and the "assignment landed, dispatch did not" message exists for exactly that.
- **The race fix is not reproduced under a real concurrent merge.** It is guarded rather than proven: worst case the loop exhausts and says so.

23/23 local, green in CI, mutations re-run against the new arms.

PR #224, opened 2026-07-26
task: measure attribution on MAIN, not on a branch ref that gets deleted (DIVE-2120)

lodar:
Found by olivia closing DIVE-2051, the acceptance test for DIVE-2101.

Both DIVE-2101 arms queried the API **by branch name**. Against an absent ref both 404, so a **merged-and-deleted branch is byte-identical to one that never existed**:

```
dive-2112-reject-actor-closed  (merged, then deleted)  ancestry=''  attribution=''
never-existed-branch-xyz                               ancestry=''  attribution=''
```

The task became permanently un-closeable, with a refusal advising *"land the branch"* for work already on main. Deleting on merge is **routine hygiene** — we deleted four the night DIVE-2101 shipped, including its own delivery.

## The fix removes the dependency rather than repairing it

Attribution now searches **main** for a commit whose **subject** names the ident. Every hit is by construction on main, so what ancestry measured is inherent — and **vacuity becomes structurally impossible** rather than separately guarded, since an empty branch contributes no such commit. It works retroactively on every existing binding, which the base-SHA alternative could not.

**Subject, not whole message.** Searching main *widens* the set, so a whole-message match accepts incidental references — measured live: `DIVE-2112` matched inside a 2-commit bound because the 0.16.20 *release* commit body names it.

## The bound is now visible (olivia's review condition)

Stopping **at** the bound is no longer reported as a miss. It returns `bound` and refuses under its own slug `done-ident-not-found-within-scan-bound`, saying **INCONCLUSIVE** and naming the number.

## Grading

19 assertions (was 15). Mutation-graded 4/4, each verified to land:

```
M1 bound-reads-as-miss    -> both bound arms
M2 query-branch-not-main  -> 7 arms incl. deleted-branch
M3 whole-message-match    -> body-mention arm
M4 attr-never-accepts     -> 4 acceptance arms
```

**M3 survived the first run.** Every fixture was single-line, so subject == message and nothing could tell the tightening from its absence. The body-mention arm exists *because* the mutation exposed that the tightening was ungraded.

No version bump — assigned at merge.


lodar:
## Iteration 2 — bar 3 fixed at the root: the bound is commits WALKED, not per_page requested

Re-delivering at `faed7b7`. Your defect is real and your diagnosis was exact; I took the pagination option rather than the one-line `min(n,100)`, and the reason matters.

### Why not `min(n,100)`

Clamping the comparison makes the message honest but leaves the tool broken in the way this ticket was filed about: work that landed 150 commits ago would be permanently inconclusive, and the refusal would have to say "the bound cannot be raised past 100". That is the same shape as DIVE-2120's original complaint — a task that cannot be closed for a reason unrelated to whether the work is on main.

So: **never request more than the clamp, and walk pages.** The scan stops when the ident is found, when history genuinely runs out (a page *shorter than the one asked for*), or when `n` commits have actually been walked. The clamp was the only known cause of a short page that is not exhaustion, and not asking past it removes that cause — so a short page now means what the code says it means. It also makes the refusal's advice **true**: raising the bound past 100 now really does walk further.

Cost: default `n=50` is still exactly one API call. Only raising the bound pays for more, which is the right place for the cost to land.

### Your measurement table, re-run against the new code

Same probe you used — the real `_gate_branch_ident_on_main`, token via `_gate_gh_token`, a nonexistent branch name:

| | before (yours) | now |
|---|---|---|
| n=50 DIVE-2101 | `[1]` | `[1]` |
| n=50 DIVE-9999 | `[bound]` | `[bound:50]` |
| n=100 DIVE-9999 | `[bound]` | `[bound:100]` |
| n=200 DIVE-9999 | `[0]` **false exhaustion** | `[bound:200]` |
| n=500 DIVE-9999 | `[0]` **false exhaustion** | `[bound:500]` |
| n=2 DIVE-9999 | `[bound]` | `[bound:2]` |

And the mirror, which matters just as much — the fix must not turn everything inconclusive:

- `n=5000 DIVE-9999 -> [0]` in 6s. main is **871 commits**, so the scan walked all of it and genuinely exhausted. Exhaustion is still detectable, now above one page.

### The number in the message is now the measured one

`bound:<walked>`, and the refusal prints what was walked, not what was configured. Claiming 200 having looked at 100 was the lie inside the defect, and the number is the one thing a reader acts on.

The message also now names **both** surviving explanations, which is your absent-vs-forbidden point applied one level up: a bound-hit cannot distinguish "landed earlier" from "nothing ever named it", so advertising only the first sends an empty-branch author off to raise a bound that can never help them.

### Your second finding — the anti-vacuity arm was unreachable

Confirmed and fixed by the same change. At the default bound on an 871-commit main the scan always hit the bound, so `_attr` was never `0` and `[[ $_anc == 1 && $_attr == 0 ]]` never fired; it passed only because every fixture was smaller than one page — a fixture-shaped pass, exactly as you called it. With exhaustion detectable, the arm is reachable again, and **ANC-11** grades it on a 150-commit history with the bound at 200, i.e. above one page rather than under it.

### The stub was lying, and that is why every arm missed this

The gh stub honoured `per_page` literally, so a fixture asking for 200 got 200 rows. **No fixture could reproduce your bug** — the harness was scoring the fix green against a world where the bug is impossible. The stub now models GitHub's clamp (caps at 100) and honours `page`. Your measured `50->50, 100->100, 200->100, 500->100` is the fidelity claim it encodes.

New arms:

- **ANC-10** — 300-commit main, `n=200`: must be a **bound-hit**, not a miss. Also asserts the message reports 200 *walked*, and that it names both explanations.
- **ANC-11** — 150-commit main, `n=200`: page 2 comes back short, exhaustion detected, the **vacuity** slug fires.
- **ANC-12** — a hit on **page 2** still closes, so the fix does not trade a false miss for a false refusal.
- plus: page 2 is actually requested, and **no request ever exceeds 100**.

26/26 green. Mutation-graded, both anchors asserted present before running:

| mutation | result |
|---|---|
| `(( count < per ))` → `(( count < n ))` (your original comparison) | **5 arms RED** |
| drop the `per > 100` cap | **6 arms RED** (including "a request asked for more than the clamp") |

### The open question in the ticket body, answered

The body flagged six blocked tasks with a `Branch:` binding and no remote ref, explicitly *not* established as casualties. Measured, `n=50` vs `n=2000`:

| ident | n=50 | n=2000 |
|---|---|---|
| DIVE-1920 | `[1]` | `[1]` |
| DIVE-2084 | `[1]` | `[1]` |
| DIVE-1952 | `[bound:50]` | `[0]` |
| DIVE-1956 | `[bound:50]` | `[0]` |
| DIVE-2044 | `[bound:50]` | `[0]` |
| DIVE-1959 | `[bound:50]` | `[0]` |

So **two of the six** (1920, 2084) are found on main and are unblocked by this path. The other four are **genuine misses even after walking the entire history** — nothing on main names them in a commit subject, so they are blocked for a different reason and this fix does not and should not unstick them. The empirical hint was half right, and the half that was wrong would have been easy to report as a win.

### What I am least sure of

That the clamp is the **only** cause of a short page. It is the only one documented and the only one either of us has measured, and the code says "the only known cause" rather than asserting more than that. If GitHub ever returns fewer than `per_page` for some other reason, this returns a false `0` again — same shape, new cause. Worth knowing that the guarantee rests on that one assumption.

PR #234, opened 2026-07-27
fix(agent): agent list carries the measured sudo grant, not just the label (DIVE-2088)

lodar:
DIVE-2079 made `agent info` honest but left `list` out of scope, so the **survey** surface kept printing the stored `isolation` label with no measurement beside it. The fixed command was the one you run once you already suspect something; the unfixed one was the command that would have told you to suspect it.

Reuses `agent_sudo_grant` / `isolation_implied_by_grant` verbatim — deliberately the same instrument as `info`, since a cheaper-but-weaker second measurement under a friendlier name would recreate the exact defect 2079 removed.

- `--json` gains the same `sudo` object shape `info` reports.
- Table gains a SUDO column: `!` for label/grant divergence, `+` for unrecognised extra entries, detail left to `info`.
- Unmeasurable prints `unknown` and never falls back to the label.

## Verification

Maker dev3, independently re-verified by olivia in her own clone (she bundled the branch out of dev3's worktree and re-derived every number rather than taking the report):

- Bundle integrity by **committed** sha, not a local build: `git show <c>:5dive | sha256sum` == `<c>:5dive.sha256` (fe8a84b5).
- `tests/agent_list_sudo_unit.sh` — 33 passed / 0 failed with the change; 27 red on the pre-fix tree, with the mutation asserted as landed before the red was trusted. Non-vacuous against the **current** base, not the one it was written on.
- Neighbours green on the rebased tree: `agent_sudo_grant_unit` 37/0, `agent_isolation_unit` 44/0.
- The new `isnt` negative-assertion helper was graded by mutation rather than by reading — forcing `_sg_implied=admin` reds it with want/got.

Rebased with `--onto` onto c47d08f; the prior remote tip was orphaned when DIVE-2079 squash-merged as 88a7678 (#207). Unversioned at main's 0.16.27 — the DIVE-2118 performer assigns at merge.

## Known residual, filed separately

Only root can measure a peer: `/etc/sudoers.d` is `0700 root`, so the read is caller-scoped and `agent info <peer>` prints `sudo: unknown` for every non-root caller (live now at installed 0.16.24). Tracked as DIVE-2135. Does not block this — `unknown` beats a false `admin`.


lodar:
Rebased onto main @ b27f57b (0.16.28) after #233 landed — the conflict was the generated `5dive` bundle serialising the merge queue (DIVE-2091), not the change. Bundle rebuilt with `./build.sh` rather than accepting git's auto-merge of the artifact; committed `5dive` hashes to the committed `5dive.sha256` (7ca1a3b8).

Re-ran the full set on the rebased tree, matching olivia's numbers exactly: `agent_list_sudo_unit` 33/0, `agent_sudo_grant_unit` 37/0, `agent_isolation_unit` 44/0. Auto-merge armed.

PR #236, opened 2026-07-27
fix(agent): privileged sudo-grant read so the measurement can see a PEER (DIVE-2135)

lodar:
Maker dev3, reviewed and passed by olivia (iteration 1). Pushed by main — dev3 holds no gh credentials.

`sudo_grant_lines` gains a last-resort `sudo -n` read, so a full-sudo caller (root, `claude`, the dashboard exec tunnel) can finally measure a **peer's** sudo grant instead of reading `unknown` for everyone. A binary-scoped agent caller still reports `unknown`, by design. `agent list` pays ONE batched privileged read for the whole fleet; `agent info` stays per-row.

Sits directly on `a483c64` (current main), so no rebase and no bundle conflict.

## Verification

- New harness `tests/agent_sudo_fallback_unit.sh`: 44/0 on branch, 22/22 **graded** failures on pristine main.
- Provenance beyond sha-parity: a clean extraction of `25173b0` + `./build.sh` reproduces the bundle byte-identically at `cde7bc4c9718e586`. Independently confirmed by me — sha-parity alone compares two outputs of the same build and cannot see a hand-edit.
- olivia ran the granted path live and confirmed the denied path is non-vacuous: the same rows that read `root-all` as a privileged caller read `unknown` when the survey runs as a binary-scoped agent. Exec count measured at 1 per fleet survey.

## ⚠️ Open discrepancy this PR exists to settle

Maker and verifier both report `agent_sudo_grant_unit` at **37/37** on this branch. I measure **30 passed / 7 failed**, apples-to-apples from clean `git archive` extractions of the same repo on the same host:

```
main     a483c64  ->  37 passed, 0 failed
branch   25173b0  ->  30 passed, 7 failed
```

The crux failure:

```
FAIL unreadable dir -> unknown, NOT none
   want: unknown|-|0
   got:  none|-|0
```

That is **absent-vs-forbidden**: an unreadable sudoers dir reporting `none` ("this agent has no sudo") rather than `unknown` ("I could not measure") — the exact property DIVE-2079/2088 established, and the one this PR's own description promises to preserve.

**Most likely my environment is the confound**, not the code: I run with full sudo, so the new privileged read *succeeds* against the fixture's "unreadable" dir and `none` becomes the honest answer — meaning the assertion encodes a pre-fallback world. If so the harness is stale and should be updated as part of this change, not left to fail.

Either way I am not merging on my own reading. CI runs this suite in a clean environment with a different privilege profile, so **`test` is the neutral referee** — and it is a required context, so it decides this rather than any of us.


lodar:
**CI is RED on `test` and `test-installed-host`, which settles the discrepancy this PR was opened to settle — and olivia predicted it before it ran.**

I had framed CI as a neutral referee. That was wrong, and her correction is the useful part: `unit-tests.yml` is `runs-on: ubuntu-latest` with no `container:`, so the job runs as `runner` — **non-root with passwordless sudo**, i.e. the *privileged* population. CI is another privileged caller, not a tiebreak. Root would have passed, because the code takes an `id -u == 0` early return and never attempts the privileged read at all.

So the three readings are consistent, and nobody misreported anything:

| caller | population | result |
|---|---|---|
| dev3 (cli-scoped) | denied | 37 / 0 |
| olivia as `agent-quinn` | denied | 37 / 0 |
| main as `claude` | granted | 30 / 7 |
| CI as `runner` | granted | **red** |

**Verdict: fixture defect, not a code defect — and still merge-blocking.** `agent_sudo_grant_unit`'s "unreadable dir -> unknown" now asks a *compound* question — is the dir unreadable to the caller **and** can the caller escalate past that — and the second term is the runner's identity. The assertion was sound when written; the privileged fallback made it underspecified.

Fix, back with dev3:

1. Stub `sudoers_privileged_dump` in the old suite — the seam already exists, dev3 built it for exactly this and used it in the new suite; the old one was never updated.
2. Split the assertion into its two real arms: privileged read **denied** -> `unknown`; privileged read **granted**, dump complete, no file -> `none`.
3. Restore the `(B) both guards reverted` arm under the stub. **That one first** — it is the suite's own non-vacuity control, and under a privileged runner the suite loses the ability to prove it can detect the defect it exists to detect. A mutation control that fails quietly is how a suite starts grading nothing while still reporting a number.

Everything else held under independent check: provenance rebuild reproduces `cde7bc4c9718e586` byte-identically, the bundle carries the feature, new harness 44/0, `agent_list_sudo_unit` 33/0, branch sits directly on current main.

Worth noting for the record: **this is the branch-protection added ~an hour ago doing its job.** Without a required `test` context this would have been mergeable on open.

PR #242, opened 2026-07-27
test(gate): fixture identities must not collide with real agent names (DIVE-2163)

lodar:
Diagnosed by **dev**, who traced it with `bash -x` rather than re-running and hoping.

`gate_verifier_route_unit.sh` hardcoded `maker_agent='dev'` / `verifier='main'` — **both live agent names**. `cmd_task_reject`'s DIVE-2112 *reject-by-maker* guard derives the actor from the real unix login:

```bash
_rj_actor=$(task_actor "")
if [[ -n "$maker" && "$_rj_actor" == "$maker" ]]; then policy_refuse E_CONFLICT reject-by-maker
```

So the harness's verdict depended on **who ran it**:

| runner | actor | role in fixture | result |
|---|---|---|---|
| agent-main, CI (`runner`), anyone else | not `dev` | verifier | **7 passed, exit 0** |
| **agent-dev** | `dev` | **maker** | refused → **exit 5 at assertion 6 of 7, no summary** |

**The guard is correct. The fixture was wrong to borrow a real identity.** Renamed to `fixture-maker` / `fixture-verifier`, which no agent holds, so no caller can ever be mistaken for a party in the fixture.

## Verified in the population that actually failed

Not just as the author — that would have proved nothing, since the author's population always passed:

```
as agent-main : 7 passed, 0 failed
as agent-dev  : 7 passed, 0 failed    <- previously exit 5 at assertion 6
```

## The larger defect is NOT fixed here

`policy_refuse` **exits**, so the harness *stopped* rather than grading assertion 6 — no summary, no FAIL line, exit 5. A test that stops is not a test that fails, and the `harness-verdict` probe cannot tell the difference: the harness never reaches its own `[[ $FAIL == 0 ]]`, so it reads **not-reached** rather than broken.

CI is green only because its runner sits in the passing population. Filed separately as **DIVE-2162**; this PR deliberately fixes only the fixture.

My own first hypothesis was that this was privilege-shaped (sudo). dev corrected it: right class — *a test whose verdict depends on who runs it*, same as DIVE-2135 — but keyed on **the fixture name**, not on privilege.

---

**Ticket-reference correction (second time tonight, recorded rather than quietly fixed).**

This PR's commit and title originally cited `DIVE-2162` — an ident I used **before filing anything**. The board then allocated `DIVE-2162` to the *other* half (the `policy_refuse` exit defect). This work is **DIVE-2163**; the commit has been rewritten to cite it (tree byte-identical, message only).

I wrote the rule *"cite a ticket you created, never one you assumed was free"* after making this exact mistake on PR #240 earlier tonight — and broke it inside the hour. That is direct evidence for DIVE-2146's thesis: **a rule you must remember at the moment of action is not a control.** The durable fix is to file the task first and let the board hand you the ident.

### DIFF
--- a/tests/gate_verifier_route_unit.sh
+++ b/tests/gate_verifier_route_unit.sh
@@ -44,45 +44,45 @@ route_sent()  { local i n; for i in $(seq 1 10); do [[ -s "$ROUTE_FILE" ]] && br
 route_last()  { local i; for i in $(seq 1 10); do [[ -s "$ROUTE_FILE" ]] && break; sleep 0.05; done; tail -n1 "$ROUTE_FILE" 2>/dev/null; }
 
 # Org chart: main is the lone root/coordinator; dev reports to main.
-db "INSERT INTO agents_org(name,reports_to,role) VALUES('main',NULL,'coordinator');"
-db "INSERT INTO agents_org(name,reports_to,role) VALUES('dev','main','builder');"
+db "INSERT INTO agents_org(name,reports_to,role) VALUES('fixture-verifier',NULL,'coordinator');"
+db "INSERT INTO agents_org(name,reports_to,role) VALUES('fixture-maker','fixture-verifier','builder');"
 
-# A live maker→verifier loop task: maker=dev, verifier=main, dev holds it.
+# A live maker→verifier loop task: maker=fixture-maker, verifier=fixture-verifier, dev holds it.
 seed_loop() {
   db "INSERT INTO tasks(ident,title,status,created_by,assignee,verifier,maker_agent,iteration,max_iterations)
-      VALUES('$1','loop task','todo','dev','dev','main','dev',1,5);"
+      VALUES('$1','loop task','todo','fixture-maker','fixture-maker','fixture-verifier','fixture-maker',1,5);"
 }
 
 # ---- 1. maker's decision gate routes to the verifier agent, not the human ----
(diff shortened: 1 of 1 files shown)


lodar:
Superseded by #257 (DIVE-2190), merged as b5ba94c, which fixes the same defect better.

This PR renamed the fixtures to fixture-maker/fixture-verifier so no real agent could collide. #257 instead PINS the caller identity — it wraps the real task_actor and forces FIXTURE_ACTOR=fixture-runner — so who runs the suite stops being an input at all, the way STATE_DIR already is.

That is the stronger fix and I want to be explicit that mine was the weaker one. DIVE-2163's body claimed the rename 'removes the CLASS rather than the instance'. It does not: a rename only removes collisions you can enumerate, and identity is ambient, so any future guard reading an identity the rename did not anticipate is caller-dependent again. Pinning removes the whole variable, and wrapping the real resolver keeps the fixture from drifting from production.

Verified on origin/main today in the population that actually failed: 9 passed / 0 failed, rc 0, as agent-main, as USER=agent-dev SUDO_USER=agent-dev (the identity that used to exit 5 at assertion 6), and as agent-olivia.

Also closing because this branch no longer applies — its base is 5570a237 against main at 25d9f718, and it conflicts on the one file both changes touch. Nothing here is worth rebasing.

PR #253, opened 2026-07-27
cli: generate the bundle at TAG time, stop carrying it on main (DIVE-2091)

lodar:
The built 47k-line bundle and its checksum were committed on main, so **every merge invalidated them on every open PR**. Measured cost in one day: three forced rebases, two on an outside contributor, one of whom sat blocked eight hours on a generated file he should never have touched.

Possible only because DIVE-2144 already moved publishing off main HEAD. `install.sh` fetches `$REPO/5dive` at the newest **tag's** sha — so the bundle must exist in the tag's tree and nowhere else. `release-cut.yml` now builds it and tags a **release commit** parented on main's tip; pushing the tag carries the objects. Main stays clean.

**Fail closed.** A tag cut with no bundle is not a loud failure — the tag exists, the release page renders, boxes keep their old version and new installs die on the download. Five guards run *before* the tag exists: build failed, no bundle, bundle vs its own `.sha256`, not valid bash, and `FIVE_VERSION` disagreeing with the tag.

**Enumerating dependents took three passes and each found more.** Patterns I expected → `bundle-drift` only. What *reads* it → 4 more call sites (`unit-tests` ×3, `install-smoke`) that use `./5dive` and never build it. What *writes* it → the whole version pipeline: `version-assign-push-loop` would fail `git add` on an ignored path (the DIVE-2142 shape), `version-assign.sh` would report UNDETERMINED forever, and `version-bump-guard` would become a permanently-passing check. All three confirmed by running them, then re-keyed onto `git diff --quiet BASE NEW -- src build.sh` — which is what the checksum was always a proxy for, as that guard's own header says.

`bundle-drift`'s old `git diff --exit-code 5dive` passes **vacuously** on an untracked file. Replaced with two that can fail: the pair must stay untracked, and the build must be reproducible.

**Credit to dev**: build-twice-and-compare is a *reproducibility* check, not a *provenance* one — `build.sh` is itself a source file, so a stale recipe emits a wrong bundle deterministically and regenerates a matching checksum (reproduced: 47,779 lines vs 48,024, `capability.sh` absent while its call sites ship). Tag-time generation closes that by building in a single detached checkout where recipe and source cannot diverge. Documented inline.

Rebased onto `7d96931`; verified DIVE-2089 is intact and all suites re-run *after* the rebase.


lodar:
**Evidence from 2026-07-29, added here because DIVE-2091 and DIVE-2185 are both closed and their bodies are frozen.**

Overnight cost of the committed bundle, measured rather than estimated: **seven rebase-and-regenerate cycles in one session** — #287, #288, #289, #291, #292, #293 (twice), #294 — hitting five different authors (dev, dev2, dev3, main2, and me as merger). Every one was the identical conflict in `5dive` + `5dive.sha256` with **zero source overlap**; the branches touched `cmd_council`, `cmd_heartbeat`, `cmd_skill` and `cmd_task` respectively.

**Two costs beyond the rebase labour, neither of which is time:**

**1. A conflicting PR yields no test signal at all — not a red.** (dev's finding.) When #294 first pushed conflicted, four required workflows simply never ran and the PR sat looking merely blocked. Absent checks and failing checks are different states that present almost identically, so a conflict silently removes CI coverage until someone notices and rebases. Anyone resolving the conflict and merging without re-running would merge genuinely untested code. This is the same not-reached-vs-pass collapse we spent the night finding in other systems — here at the PR level.

**2. It penalises rigorous verification more than sloppy verification.** A grade is pinned to a SHA. Every merge invalidates every other open PR's tip, so a verifier who insists on measuring directly at the tip that will merge has their evidence expire faster than they can report it — twice in one night on DIVE-2257 alone. A verifier who accepts a proxy is unaffected. That incentive runs exactly backwards.

Serialising every branch through two generated artifacts means whoever merges last always rebases, by construction. That is not a discipline problem and more careful rebasing does not touch it.

lodar:
Superseded by #313, which is the same DIVE-2091 change re-cut on a fresher base. This branch also fails `docker-install` and conflicts with main. Closing to leave one queue entry for the change; the branch stays if anything needs recovering from it.

PR #268, opened 2026-07-28
fix(heartbeat): an unmeasurable tier no longer disables the privilege-escalation guard (DIVE-2213)

lodar:
Second instance of the DIVE-2210 shape, at a **decision** site rather than a display one.

## The defect

DIVE-1065 refuses to auto-drive a higher-tier agent from a lower-tier creator's task. It read both tiers as

```bash
_ctier=$(jq -r --arg n "$_cby" '.agents[$n].isolation // empty' <<<"$reg" 2>/dev/null)
_cr=$(_hb_tier_rank "${_ctier:-}")          # "" -> 0
if (( _cr > 0 && _ar > 0 && _cr < _ar )); then HOLD; fi
```

`2>/dev/null` discards *why* the lookup failed, `// empty` discards *that* it failed, and rank 0 folds the residue in with "creator is a human". So a lookup that never happened did not hold the task, it **disabled the check**, and the heartbeat auto-ran the work.

## The ticket's binary is a false binary

The ticket asked: hold everything unmeasured (fail closed, may stall the fleet) or wake with a loud log (fail open, no longer silent)? Neither. Rank 0 was hiding two populations that want opposite policies:

- **measured, no tier** - the creator is not a registered agent (a human, an external filer). The majority of the board. Falling through is DIVE-1065's *intent*; holding here stalls every human-filed task fleet-wide, which is exactly the cost the ticket feared.
- **not measured** - registry absent/unreadable/unparsable, jq errored, or a **registered** agent whose `isolation` is missing/malformed. No basis to rank either side.

Only the second holds now. A healthy registry never produces it, so this cannot stall the fleet in steady state, and every hold names its own cause in the tick log.

`agent_tier()` + `tier_unmeasured()` (`src/lib/registry.sh`) draw that line. They are deliberately **not** `envelope_tier()`, which reports `unknown:unregistered` for both populations on purpose - a wire format has no decision to make, and `tests/envelope_tier_provenance_unit.sh` asserts that fold. The suite here asserts `envelope_tier()` is **byte-identical to origin/main**, so DIVE-2210's shipped wire format does not move.

## Measurement

The guard block is extracted **verbatim** from this tree *and* from `origin/main` and driven through the same nine causes by the same harness - the pre-fix code is not retyped, so the RED cannot silently agree with my reading of it.

```
distinct decisions across 9 causes: OLD=2  NEW=3
OLD: WAKE WAKE WAKE WAKE WAKE WAKE WAKE HOLD:escalation WAKE
NEW: WAKE HOLD HOLD HOLD HOLD HOLD HOLD HOLD:escalation WAKE
```

The pre-fix block auto-ran on **6 of 6** unmeasured causes while still holding the one real escalation and still waking on a human creator, i.e. it looked like a working guard. 41 assertions, 0 failures.

**Mutation-graded (5/5 killed, on the committed tree):** re-fold `no-tier` as measured; restore the `envelope_tier` fold; over-hold on `unregistered` (the fleet-stall direction); return empty on a failed read; drop the unmeasured hold. The fourth drives `guard/registry-absent -> WAKE`, the ticket's own defect reproduced inside the suite.

## Reachability, stated rather than overclaimed

Pre-fix, a whole-registry failure could **not** reach this guard: the wake loop enumerates agents from the same `$reg` blob, so a failed read yielded zero agents. Reachable pre-fix were the three that leave the registry loadable - a registered-but-untiered creator, a malformed tier, a jq failure. The other three become reachable *after* this change, because `agent_tier()` re-reads at decision time and so also catches a registry that dies mid-tick. The OLD=WAKE column is a property of the code, not six live incidents.

## Third instance, display-only

`task show`'s `created_by_tier` line was printed only on a non-empty lookup, so a failed lookup and a genuine absence both rendered as no line. Always printed now, in three distinguishable states. Checked for parsers first: `origin/main` across 5dive-cli / api / app / plugins / mcp has no consumer other than the emitting site, and the machine path is `--json`, which never carried it.

Observed against the live board, new build vs shipped 0.16.35:

```
DIVE-1574  created_by = agent-main
  new:      created_by_tier = none (creator is not a registered agent)
  shipped:  (no line)
```

Which surfaced something real: **9 tasks on the live board are created by `agent-main`, and that is not a registry key** - the key is `main`. An agent creator currently reads as "not an agent" through an alias. Harmless in this instance (main is admin, the safe direction), but it is the renamed-agent hole in live data, and it is now visible instead of silent.

## Not done here

No live tick was driven. This proves which branch the guard takes, not that a tick reaches it; reachability is reported above, not asserted. `FIVE_VERSION` is not bumped (assigned at merge, per CONTRIBUTING).

Generated with [Claude Code](https://claude.com/claude-code)


lodar:
Sweep result, per DIVE-2210's parting instruction to find the next instance — recorded here rather than folded in.

Re-running the `2>/dev/null` + `// empty` grep on this tree is mostly benign, but it surfaced a **second family the registry-shaped grep could not see**: `AGENT_ISOLATION` read out of the agent **env file**, an independent second source of truth for the same fact that nothing reconciles against the registry. Filed as **DIVE-2218**, not fixed here.

- `cmd_agent_teambot.sh:1242` — `iso="${iso:-admin}"` then `write_agent_env ... "$iso"`. It does not just read a wrong value, it **manufactures the top tier and persists it**, so a transient read failure becomes a durable `admin` label on disk. Strictly worse than any read-only instance.
- `cmd_skill.sh:106` — same idiom, but the direction **degrades** rather than escalates (a sandboxed agent reads as not-sandboxed, the install runs as the wrong user and fails). Recorded as functional, not privilege. Three hits is not three escalations.
- `cmd_agent_pairing.sh:812` — **not a defect, and it is the model for the fix.** Defaults to `standard` with a comment explaining the fail-safe direction (DIVE-1571).

Worth noting for anyone reviewing the scope line: the correct pattern was already in this codebase one file over, and the second family was invisible to the sweep the first ticket prescribed — `grep '// empty'` finds jq lookups, not `sed -n 's/^X=//p'`. A sweep is scoped by the idiom you already know.

lodar:
Closing the one loose end in the `task show` half of this PR. The body says "checked for parsers first"; here is the actual enumeration, because a grep for the field NAME only rules out consumers that mention it.

Consumers of `task show` on `origin/main`:

- `5dive-plugins/plugins/telegram/server.ts` — **4 of 4 call sites pass `--json`** (3797, 3965, 4131, 5126), and `tna.ts`'s `TnaGate` is a structured object of the JSON columns. The `--json` branch never carried `created_by_tier` and is untouched here.
- `5dive-mcp/src/index.js:126` — returns `["task", "show", id]` with **no `--json`**, so it is the one live consumer of the TEXT path. Its output is handed to a model as an MCP tool result, not to a parser, so an added line is additive rather than breaking — and it is the consumer that most benefits from being able to tell "no tier" from "not measured".
- `app/.../task-row.tsx` — comments referencing `task show`'s field set; the dashboard reads the API, not the text.

So: no positional or line-count parser exists on the text path. Worth stating the method rather than the conclusion — grepping `created_by_tier` proves nothing about a consumer that splits the output generically, which is why this went looking for `task show` INVOCATIONS instead.

PR #281, opened 2026-07-28
fix(version-assign): measure the ABSOLUTE property, not a delta an adjacent commit can erase (DIVE-2230)

lodar:
## What

`version-assign` compared main's bundle to the bundle at the **previous commit**, which asks *"did the bundle move in THIS push"*. The invariant it exists to defend is *"does main's bundle match the **last assigned version**"*. Those two agree exactly until an assignment fails to land — after that the debt is carried by no one and **erased by the next unrelated commit**, permanently and silently.

Measured on this repo: run **#268** (`a2e2009`) printed `ASSIGNMENT OWED ... next = 0.16.36` and died on a branch-protection push rejection — loud and correct. Run **#271** (`0e37bf7`), a workflow-only push, then printed `no assignment needed — the bundle is unchanged since a2e2009` and went **green**, with main's bundle still unassigned. Every run after that was green too. `bundle-drift` cannot see it either: it compares bundle to SOURCE, and those agree.

This is not a bot that broke and needs a retry. It is **a check measuring a DELTA when the property is ABSOLUTE** — the same wrong-target shape as an anchor pinned to a moving ref (DIVE-2213).

## How

The base is now **derived**: the commit that set `FIVE_VERSION` to its current value, whose recorded `5dive.sha256` *is* the bundle that version shipped. The comparison answers the question at **any** commit, independent of what the adjacent one did, so a transient push failure can no longer launder itself green.

`<base-rev>` is still accepted and still validated, as a **truncated-history canary** — a shallow fetch is precisely what breaks the anchor walk — but it decides nothing. A caller cannot reintroduce the delta semantics by passing the wrong thing, which is why **`.github/workflows/version-assign.yml` needs no change**.

Two fail-opens were closed rather than moved:

- **The derived walk imports its own.** "The version never changed in visible history" is emitted identically by *this is genuinely the first version* and *the history is truncated and the assignment is below the graft*. Guessing either way restores the original defect one layer down, so it asks git whether the history is complete (`rev-parse --is-shallow-repository`) and **refuses** (exit 2) when it is not.
- **The two old "nothing owed" branches collapsed into one.** Both were proxies for the same absolute fact. If the old exemptions had survived the conversion, it would not have been absolute yet.

## The live instance, stated precisely

**main is carrying this debt right now.** `0.17.1` was assigned at `362b975` against bundle `a722b6ee`; main's tip `17eb18c` ships `7af4dd0c`. Two bundle moves landed since (#278, #279) with no assignment, every run green. Pointed at that tip this branch reports `ASSIGNMENT OWED ... next = 0.17.2`; the shipped script reports `no assignment needed`.

**This is prospective harm, not a live customer incident.** Independently checked by @agent-main: the **`v0.17.1` tag ships `a722b6ee`** — the same bundle the assignment named — and boxes install the newest tag, not main HEAD (DIVE-2144). Every published artifact is internally consistent and no install is mismatched. The exposure is that the *next* cut is either blocked by the uniqueness gate or publishes a second artifact under a name already taken.

**Do not hand-assign `0.17.2` before this lands.** The outstanding obligation on main is the only un-fixtured test case this fix will ever get for free; clearing it first leaves the new logic's first production run with nothing to find — which is the exact trap in the ticket's own writeup. @agent-main is holding the assignment until this is on main.

## Grading

The regression case is **not** "a push fails" — the broken implementation passes that, because at the failing commit the delta and the absolute check agree. It is "a push fails, **then** an unrelated non-bundle commit lands".

So arms I1–I4 are differential by construction, against a one-line mutant restoring the old comparison:

- **I0** asserts the mutation actually landed. A `sed` that matched nothing would have every arm below comparing the script to itself and reporting a confident green.
- **I1** asserts **both** implementations report the debt on the first half. A keystone arm that passes on both is not a keystone, and this makes that explicit rather than assumed.
- **I2** asserts they **disagree** on the second half. This is the regression.
- **I3** holds it across five unrelated commits, so a fix that merely widened the window fails.
- **I4** is the non-vacuity arm: once the assignment lands, later doc commits go quiet. Without it, `echo OWED; exit 0` passes I1–I3.
- **J** refuses on a genuinely shallow clone (depth 2 on purpose, so the pre-existing base check passes and the arm grades the *anchor* path).

**36/0 with the fix, 28/8 without it, same harness.** Also green: push-loop 40/0, uniqueness-gate 12/0, bump-guard 21/0, push_unit 78/0.

`scripts/` is not embedded in the bundle, so **no rebuild and no version bump** — 3 files, +194/−17, no `src/`, no `5dive`, no `5dive.sha256`.

## Known residual — DIVE-2264

`version-assign.yml` still exits 0 when `github.event.before` is unusable. Under absolute semantics that skip is a fail-open of the same family, since the question no longer needs `before` at all. Not fixed here: it lives in a `.github/workflows/` file, which `5dive push` cannot ship (DIVE-2262), and the available workaround (the box's `lodar`-scoped `gh` credential) would record a human as editing CI — declined per DIVE-2232. Filed as **DIVE-2264** rather than left in this PR body, because a residual that lives only in a PR body is not tracked.

## Unverified connection, attributed

@agent-olivia notes this may be the silent half of **DIVE-2243** (the host is on `0.16.32` while customers are on `0.17.1`), on the mechanism that `release-cut` never tagged because it counted its own in-progress run, publishing resolves from the newest tag, and `install.sh` has no version comparison. **I did not verify that mechanism** and am not asserting it. One datum that complicates it and is worth someone checking: `v0.17.0` and `v0.17.1` **do** exist as tags, while `0.16.33`–`0.16.36` have none.

Reviewed for scope and shape by @agent-olivia (explicitly against a non-empty diff); blast radius corrected by @agent-main. Neither ran the suite — behaviour is CI's to establish.

Closes DIVE-2230.

### DIFF
--- a/scripts/version-assign-push-loop.sh
+++ b/scripts/version-assign-push-loop.sh
@@ -34,11 +34,25 @@ git config user.email 'markounik@gmail.com'
 # scenario this job exists for. So: refetch, recompute against the moved tip, retry,
 # and fail loudly.
 #
-# BASE stays github.event.before across retries, deliberately. If a previous run
-# already assigned, the version HAS moved relative to that base, so the script reports
-# "already moved" and we stop — one version per batch, which is the batching lodar
-# wants. If nobody assigned, we assign once for the whole batch. Both are correct;
-# neither double-bumps.
+# BASE IS NOW INERT (DIVE-2230), and the batching survives it. version-assign.sh no
+# longer decides against the base at all — it compares main's bundle to the bundle
+# recorded at the commit that ASSIGNED the current version, which is an absolute
+# property of the tree and not a delta anyone can launder away. $BEFORE is still
+# passed, and still validated by the script, purely as a truncated-history canary: an
+# unresolvable base means a shallow fetch, and a shallow fetch is what would break the
+# anchor walk.
+#
+# One version per batch still holds, and now holds for a BETTER reason. The loop
+# resets to the freshly-fetched tip before computing, so the assignment it makes
+# covers every merge landed so far. A sibling run then finds the anchor's bundle equal
+# to the tip's and reports nothing owed — not because "the version moved relative to
+# my base", but because main genuinely ships the bundle its version claims.
+#
+# The one deliberate behaviour change: if a further merge lands AFTER that assignment
+# and moves the bundle again, this now assigns a second version instead of stopping.
+# That merge is a real uncovered bundle, and the shared-checkout updater is
+# VERSION-triggered (DIVE-2065) — declining to bump it is not batching, it is the
+# DIVE-2230 debt with a shorter fuse. The tag freeze is on tags; bumps are cheap.
 assigned=""
 for (( try=1; try<=ATTEMPTS; try++ )); do
   git fetch -q origin main
--- a/scripts/version-assign.sh
+++ b/scripts/version-assign.sh
@@ -22,7 +22,37 @@
 # installed-bundle-matches-its-commit selfcheck honest). So this script bumps,
 # and must NEVER create a tag or a GitHub release.
 #
+# DIVE-2230: THE QUESTION IS ABSOLUTE, SO THE BASE IS DERIVED, NEVER PASSED.
+#
+# This script used to compare the bundle at <new-rev> against <base-rev> — the
+# PREVIOUS COMMIT. That asks "did the bundle move in THIS push". The invariant it
+# exists to defend is "does main's bundle match the LAST ASSIGNED VERSION". Those
+# two agree exactly until an assignment fails to land. After that the debt is
+# carried by no one and ERASED by the next unrelated commit, silently and for good.
+#
+# Measured 2026-07-28, three artifacts: run #268 (a2e2009) printed "ASSIGNMENT
+# OWED ... next = 0.16.36" and died on a branch-protection push rejection — loud
+# and correct. Run #271 (0e37bf7), a workflow-only push, then printed "no
+# assignment needed — the bundle is unchanged since a2e2009" and went GREEN, with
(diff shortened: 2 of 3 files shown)


lodar:
Resolving the **unverified connection** flagged in the last section, since the datum that complicates it turns out to have a clean answer.

Both statements are true and they do not conflict — the two `0.17.x` tags have **different provenance**:

| release | published | by |
|---|---|---|
| `v0.17.0` | `2026-07-28T06:58:33Z` | `lodar` (hand-cut) |
| `v0.17.1` | `2026-07-28T09:22:44Z` | `github-actions[bot]` |

The `release-cut` self-deadlock fix (#275, `280d9e6`) merged at **`09:03:16Z`** — between them.

So `v0.17.0` predates the fix and was **not** produced by `release-cut`; `v0.17.1` is the first and so far only tag the automation has ever cut, 19 minutes after the fix landed. `0.16.33`–`0.16.36` have no tags at all, which is consistent with the automation being deadlocked across that whole window.

The existence of two `0.17.x` tags therefore does not refute the "never tagged" mechanism — it is one hand-cut release plus the first automated one after the deadlock was removed.

**One caveat, and it is load-bearing rather than pedantic:** `published_by = lodar` identifies the *credential*, not the actor. Per DIVE-2232 that field carries no information about whether a human or an agent performed the cut, so "hand-cut" here means "not cut by `release-cut`" — it does not establish who did it.

Method note: my first attempt at this read `git log -1 --format=%ci <tag>`, which returns the **commit** date of the tagged commit, not when the tag was created. That inverted the ordering and would have produced the opposite conclusion. Annotated-tag `taggerdate` and the Releases API agree with each other; the commit date answers a different question.

PR #299, opened 2026-07-29
feat(update): alarm on a fleet that stopped updating, and make --check agree with the installer on "latest" (DIVE-2287)

lodar:
Closes the two findings on DIVE-2287 — the work half of the residue recorded in `community/wiki/an-upgrade-that-cannot-compare-versions-dive2243.md`.

## 1. A freeze alarm that takes no comparand

Every staleness signal we ship — `update --check`, the supervisor `cliStale` probe, the dashboard maintenance tile — is the same comparison `installed < latest` wearing several faces. They look independent; they share one failure mode:

> When the **release process** stops, `latest` stops moving too. Every box compares equal, `behind` is false fleet-wide, and every face reports green in unison about the one thing none of them can see.

DIVE-2243's monotonicity guard sharpened this rather than fixing it: it converted a silent fleet-wide downgrade into a loud refusal **plus a frozen fleet**, and nothing alarms on a frozen fleet.

So the new reading is **absolute** — *has the version this box RUNS changed, ever, within 7 days* — which stays answerable with the cutter dead, the tag frozen, the probe offline and `--check` outright wrong. `_cli_freeze_observe` runs from the supervisor tick **before** the staleness probe and outside every early return in it; each of those returns is a case where the comparison could not be made, which is exactly when the absolute reading is the only signal left.

Surfaced in the supervisor summary line, `supervisor --json` (`cli.frozen`), and `update --check`.

**Not built on the installed binary's mtime**, which is the tempting baseline and wrong in the silent direction: `refresh_managed_files()` ends in an unconditional `mv -f "$_bundle_tmp" "$BIN_DIR/5dive"` with no already-current branch, so the nightly rewrites the binary every night whether or not the version moved. mtime measures the last **attempt** and never the last **change** — the two things this alarm exists to tell apart — so an mtime reading would say "moved yesterday" forever: a monitor that cannot fire, occupying the slot. The test fails the `stat`-based simplification specifically.

An absent record resolves to **UNKNOWN, never fresh**: "first pass on a new box" and "record wiped" emit the same absence (DIVE-2230's rule, one layer down).

## 2. `--check` and the installer now agree on what "latest" means

Measured on the council-demo box:

```
5dive update --check   ->  CLI 0.16.32 is behind (latest 0.17.2)
sudo 5dive self-update ->  5dive upgraded: 0.16.32 -> 0.17.1
```

The probe read `main`; `install.sh` installs the newest **tag**. Both correct in isolation; together they tell an operator they are behind a version that does not exist as a tag and no installer can deliver — permanently. With the monotonicity guard it compounds: a box above the newest tag is refused every upgrade while `--check` insists it is behind, and both messages are right.

There is one question — *what would the installer give me* — so there is one place the answer may come from. The probe now resolves the newest `vX.Y.Z` tag through the same three rungs as `resolve_gh_tag` (git ls-remote → tags.atom → tags API) and **fails closed** when none resolves, which is what the installer already does. **No fallback to main** — falling back *is* the defect.

Fetch by tag **name**, not the sha `ls-remote` prints: 97 of our tags are annotated and raw.githubusercontent does not serve a tag object. A tag is also immutable, which closes DIVE-2042's split-generation window for the same reason pinning `main`→sha did, one round trip cheaper.

**A box ABOVE the newest release is now its own state**, not the `up to date` else-branch — that is the state DIVE-2243's guard refuses every upgrade from, and spelling it "up to date" is how an operator ends up holding two correct contradictory messages with no thread to pull.

`behind` / `stale` keep their exact prior meaning; `ahead` / `frozen` are additive so existing callers read the same fields.

## Verification

- `tests/version_freeze_observer_unit.sh` — **19/0**, new. Observer block extracted verbatim from `src/`; `now` injected so no case reads the wall clock. Threshold **mutation-graded** for non-vacuity (widening the window to 999d must flip the 8d case to `moving`, and the mutation is asserted to have landed). The mtime simplification is explicitly failed (`touch` the record; the answer must not move). Plus a structural fence — labelled as structural — on the supervisor call ordering.
- `tests/update_check_propagation_unit.sh` — **22/0**, reworked for the tag rail. Adds a stub serving **only** the tag path, so a regression to reading `main` cannot resolve at all; adds the `ahead` render cases.
- **Live on this host**, which was itself in the ahead state (installed 0.17.2, newest tag v0.17.1):

```
$ ./5dive update --check
OK — CLI 0.17.2 is AHEAD of the newest release 0.17.1 — the installer will refuse to move it (a release cut is owed)
$ ./5dive update --check --json
{"current":"0.17.2","latest":"0.17.1","behind":false,"ahead":true,"stale":false,
 "frozen":"moving","frozenAgeSec":1,"frozenDetail":"CLI 0.17.2 first observed 0d ago","source":"v0.17.1"}
```

`source: v0.17.1` is the same ref `install.sh` resolves. Before this change the probe returned `0.17.2` from `main`.

## What this does NOT close, stated rather than implied

- **Delivery stops at the board.** The freeze signal reaches `5dive supervisor` and `update --check`. It does **not** yet reach the digest or a Telegram push, and `supervisor --tick` no-ops unless the enable flag exists. DIVE-2287 asked for a signal that "reaches a human or the digest" — this is the signal, one hop short of the reach.
- **Per-box, not fleet-aggregated.** "The fleet has stopped" is still an inference a reader makes from several boxes agreeing.
- **The record is only written by callers that run.** `update --check` is unprivileged and degrades to `unknown` on a read-only `STATE_DIR`; the tick runs as root and normally can write. Where neither runs, the three-state output says "unknown" out loud rather than smoothing it into a green.
- **Deliberately not a recurring task** (a monitor implemented as one can suppress itself — `community/wiki/pile-up-is-noise-for-a-chore-and-signal-for-a-monitor.md`) and deliberately **not keyed off `in_progress`**, which became a weaker signal on 2026-07-29 per the DIVE-2244 coupling recorded on the ticket.
- **The cutter is still down.** Making the two halves agree does not cut a release; DIVE-2238 remains the root cause of all three faces.

Wiki: `community/wiki/an-alarm-that-does-not-need-a-comparand-dive2287.md` (+ index line).


lodar:
**Mutation grade on the reworked harness** — `tests/update_check_propagation_unit.sh` is a rewrite of an existing suite, so "22/0" alone does not show the new assertions discriminate. Ran it against the **pre-DIVE-2287 implementation** (`origin/main`'s `_published_cli_probe` + `cmd_update_check`, with only the new freeze block grafted in so the harness can source it):

```
12 passed, 10 failed

FAIL - mismatch message is wrong
FAIL - happy path wrong
FAIL - probe did not fetch from the tag path
FAIL - tag selection wrong
FAIL - atom fallback broken
FAIL - broken-git fallback wrong
FAIL - no-tag case did not fail closed
FAIL - ahead-of-release path did not render
FAIL - a box the installer will refuse is being reported as up to date
FAIL - ahead prose does not explain the consequence
```

The 10 reds are **exactly** the DIVE-2287 arms (tag rail + `ahead`), and the 12 that pass on **both** implementations are the DIVE-2042 propagation arms — which is the point: the split-generation behaviour this suite was written to protect is asserted to be **unchanged**, so the rework re-aimed the suite without blunting what it already graded.

Source restored and re-verified after the experiment: 22/0 and 19/0, working tree clean, rebuild reproduces the identical bundle sha (`e7584f43…`).

PR #323, opened 2026-07-29
chore: DIVE-2372 positive control — same branch, PR opened by a REAL credential (DO NOT MERGE)

lodar:
Liveness/positive-control arm for the DIVE-2372 probe. Identical head branch that GITHUB_TOKEN pushed; this PR is opened by a real credential so we can tell 'no runs because the event was suppressed' from 'no runs because nothing would ever run on this branch'. Closing immediately after the count.


lodar:
DIVE-2372 probe complete — measurement taken, closing. Not for merge.

PR #324, opened 2026-07-29
fix(gate): iteration 2 — route the refusal guard through the seam, de-vacuate T8, unmangle the comments (DIVE-2330)

lodar:
Iteration 2 of DIVE-2330. dev rejected iteration 1 with three fixable defects plus a minor; all four are addressed, and dev's `T3-T9` root cause was correct.

## (1) The T3-T9 red was the GUARD, not the seam

`cmd_task.sh`'s agent-refusal read `$EUID` **directly**, bypassing `_gate_caller_uid` — and the comment I wrote above it justified the bypass by conflating two functions:

- routing through `_gate_authenticated_actor` **would** widen the refusal (it adds the root+`SUDO_UID` branch),
- routing through `_gate_caller_uid` widens **nothing**, because the seam's entire body is `printf '%s' "$EUID"`.

The EUID-only semantics live *inside* the seam. That is the point of the seam. What the bypass cost: on any `agent-*` runner the guard fired on the **runner's own** identity, and because `cmd_task_answer` is a **sourced function**, its `fail` called `exit` and decapitated the suite at 9/18 with rc=6. The suite's outcome became a function of whose uid ran it — the DIVE-2365 host-supplied-precondition shape this branch **named in its own body and then reintroduced one function over**.

Also captured rc on T3's arm, so the first such refusal reds a named arm instead of truncating the log.

## (2) T8 was vacuous

The override did `command id -u agent-evil || printf 0`. No such user exists on any host, so it returned `0` = **root**: the arm that exists to prove an *agent* caller is refused was modelling root.

Fixed by giving the passwd **source** a function seam, `_gate_passwd_stream`, so a literal fake uid can be mapped on every host including GitHub's runner. Deliberately a **function**, not `${_GATE_PASSWD_FILE:-/etc/passwd}` — an env-settable source would be a new forgery vector in the very field this row exists to make unforgeable, which is the reasoning `_envelope_sender_fallback` already records in `cmd_agent_runtime.sh` ("NO ENV OVERRIDE for the passwd path, deliberately"). A function cannot be injected: bash imports exported functions at startup and the script's own definition executes after that import and overwrites it — the same property `_gate_is_root` and `_gate_caller_uid` already rely on. The fixture **prepends** to `/etc/passwd` so real uids still resolve.

Added a **pin-check** that aborts if the seam pair resolves to nothing (a vacuous arm prints `ok` — that is how this got through), and **T8b**, the non-agent control on an otherwise identical gate, so T8 grades a mechanism rather than a message.

## (3) The comments were mangled — the artifact this row exists to prevent

- `registry.sh`: annotation **removed, not repaired**. dev is right on the merits — it attached an identity-resolution claim to `tier=` in the `[5dive-msg]` envelope, a different mechanism this fix does not touch.
- `cmd_task.sh`: the DIVE-2099 sentence is restored (duplicated fragment and glued `#` gone); the lead-clear annotation is reindented into its block.
- `cmd_agent_runtime.sh` needed **no change**, and I verified rather than assumed: its block already reads `$EUID`, already says `tier=` is *not* the unforgeable field `registry.sh` claims, and already attributes the vector to DIVE-2330.

## (4) Minor

The forgery harness now evals `_gate_is_root` and `_gate_passwd_stream` — without the former the `SUDO_UID` branch was graded against a missing function (`command not found`, twice) — with a `declare -F` assertion so a failed eval cannot pass vacuously. A skipped arm now prints `skip` against its own counter instead of inflating the pass count.

## Verification — both runners, one pinned tree (`5c30316`)

| runner | `gate_nonce_unit` | `gate_actor_path_forgery_unit` |
|---|---|---|
| `agent-main` (uid 1004, `agent-*`) | **19 passed, 0 failed** | 5 passed, 0 failed, 0 skipped |
| `claude` (uid 1000, non-agent) | **19 passed, 0 failed** | 4 passed, 0 failed, **1 skipped** |

Identical on both, which is the property dev asked for: the outcome is no longer a function of whose uid runs it. (Iteration 1: 9/18 + exit 6 as `agent-*`, 17/18 as `claude`.)

### Negative controls — both mutations red, on both runners

| mutation | `agent-main` | `claude` |
|---|---|---|
| restore the `$EUID` bypass in the guard | T3 **FAILS** by name, 12 arms, rc=6 | **18/1** — T8 reds, because the guard stops seeing the pinned identity |
| revert the resolver to a hardcoded `< /etc/passwd` | pin-check aborts: `NOT OK - seam pin is inert` | same |

Restored to green after each.

**Residual, disclosed not discovered:** under the guard mutation as `agent-*` the run still exits 6 at 12 arms. Capturing T3(a) converts the *first* refusal into a named failure; later uncaptured calls in that file still propagate. The arm now says which mechanism broke, which is what was missing — but this file is not fully exit-proofed, and that is a bigger edit than this row should carry.

### DIFF
--- a/src/cmd_task.sh
+++ b/src/cmd_task.sh
@@ -540,7 +540,9 @@ _gate_route_reviewer() {
 #     agent id from SUDO_USER (auto_sender_from_sudo) and the human signal from a
 #     non-agent SUDO_UID (_gate_sudo_uid_nonagent), mirroring cmd_task_answer.
 #   - non-root: SUDO_* are UNTRUSTED (forgeable) — ignore them entirely and judge
-#     by the real `id -un` unix login, which a process cannot spoof.
+#     by $EUID resolved against /etc/passwd in pure bash. DIVE-2330: this line used
+#     to say `id -un` "which a process cannot spoof" — it could, because `id`
+#     resolves through the CALLER'S PATH and a shim printing any name won.
 # Prints one of: "agent <name>" | "human" | "none". Never reads --from ($from stays
 # ATTRIBUTION-only) nor $USER (also env-spoofable). The determined root-sudo residual
 # (an agent that truly sudo's then forges) is shared with the whole gate system and
@@ -554,10 +556,12 @@ _gate_withdraw_actor() {
     _gate_sudo_uid_nonagent && { printf 'human'; return; }
     printf 'none'; return
   fi
-  local idun; idun=$(id -un 2>/dev/null || echo "")
-  if [[ "$idun" == agent-* ]]; then printf 'agent %s' "${idun#agent-}"
-  elif [[ -n "$idun" ]]; then printf 'human'
-  else printf 'none'; fi
+  local _a; _a=$(_gate_uid_to_agent "$EUID")
+  if [[ -n "$_a" ]]; then printf 'agent %s' "$_a"; return; fi
+  local _n _x _u; while IFS=: read -r _n _x _u _; do
+    [[ "$_u" == "$EUID" ]] && { printf 'human'; return; }
+  done < /etc/passwd
+  printf 'none'
 }
 
 # DIVE-980: shared org-chart assignee resolution. Resolve an assignee TOKEN to a
@@ -3888,12 +3892,44 @@ _gate_floor_axis() {
 # The dashboard path is unaffected and needs no marker: shelld runs as `claude`
 # (a NON-agent user) and sends `--human`, so those answers are `human:*` — already
 # authorized everywhere `lead:*` is, and never in want of a lead stamp.
+# DIVE-2330: resolve a uid to an agent name in PURE BASH over /etc/passwd. No
+# `id`, no `getent` — both resolve through the CALLER'S PATH, and every agent can
+# set its own PATH. Measured on this host before the fix: a shim on PATH printing
+# `agent-lodar` made `id -un` return exactly that, from a process whose real uid
+# was 1004 (agent-main). Same shape as the landed DIVE-2281 fix in
+# `_envelope_sender_fallback`.
+_gate_uid_to_agent() {
+  local want="${1:-}" name _x uid
+  [[ "$want" =~ ^[0-9]+$ ]] || { printf ''; return; }
+  while IFS=: read -r name _x uid _; do
+    [[ "$uid" == "$want" ]] || continue
+    [[ "$name" == agent-* ]] || { printf ''; return; }
+    printf '%s' "${name#agent-}"; return
+  done < /etc/passwd
+  printf ''
+}
+
+# `$EUID` is a bash BUILTIN reflecting the kernel's view of this process. It is not
+# PATH-resolved and cannot be set from the environment, which is the whole property
+# an authorization check needs. SUDO_UID is consulted only when EUID is 0, exactly
+# as before: sudo writes it, the caller does not.
+# SEAM (mirrors _gate_is_root): the harness overrides THIS to exercise both
+# branches with the real resolver. An external caller cannot — invoking `5dive`
+# starts a fresh bash that defines the function itself, so there is nothing in the
+# environment to override. That is the difference between a seam and a hole: the
+# old code let a PATH shim answer, which any agent could set.
+_gate_caller_uid() { printf '%s' "$EUID"; }
+
 _gate_authenticated_actor() {
-  local u; u=$(id -un 2>/dev/null || echo '')
-  if [[ "$u" == agent-* ]]; then printf '%s' "${u#agent-}"; return; fi
-  if [[ "$(id -u 2>/dev/null || echo 1)" == "0" && -n "${SUDO_UID:-}" ]]; then
-    local su; su=$(getent passwd "$SUDO_UID" 2>/dev/null | cut -d: -f1)
-    [[ "$su" == agent-* ]] && { printf '%s' "${su#agent-}"; return; }
+  local a; a=$(_gate_uid_to_agent "$(_gate_caller_uid)")
+  if [[ -n "$a" ]]; then printf '%s' "$a"; return; fi
+  # The SUDO_UID branch stays gated on the REAL root check (_gate_is_root, itself
+  # the pre-existing seam), NOT on _gate_caller_uid. Routing it through the seam
+  # would let a harness modelling "caller is root" also unlock the SUDO_UID path
+  # and widen who resolves to an agent — a behaviour change, not a fix.
+  if _gate_is_root && [[ -n "${SUDO_UID:-}" ]]; then
+    a=$(_gate_uid_to_agent "$SUDO_UID")
+    [[ -n "$a" ]] && { printf '%s' "$a"; return; }
   fi
   printf ''
 }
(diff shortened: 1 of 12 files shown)


lodar:
## Update — a real regression I found in iteration 2, and fixed (`eed7a09`)

My own regression sweep caught this, not review. **`gate_sudo_uid_forge_unit` and `gate_t2_routed_escalate_unit` both went rc=6 on this branch while `origin/main` is green** — 8 passed / 0 failed, measured on a `main` worktree, same runner, same box, so the baseline is a real differential and not my rig.

**Cause, and it is an obligation the src fix creates rather than a surprise:** both harnesses pinned the caller identity by stubbing `id()`, because the resolver *read* `id -un`. That read is exactly the forgery this row closes, so removing it strands every pin built on it. The guard then saw the **real runner**, refused, and because `cmd_task_answer` is a sourced function its `fail` exited the harness — the same decapitation this row already fixed one function over.

Ported both onto the seams the resolver now reads: `_gate_caller_uid` for the uid, `_gate_passwd_stream` for the mapping, fixture prepended to `/etc/passwd` so real uids still resolve. `id()` stays in both files for `_gate_sudo_uid_nonagent`, which still reads `id -u` and is untouched here. Added `assert_lead_identity` as the DIVE-2330 companion to dev's DIVE-2365 `assert_agent_caller` — a pin that silently yields `''` makes `_lead_clear=0`, so the arms would grade a refusal and print `ok`.

### Verified on ONE pinned tree (`eed7a09`), BOTH runners

| harness | `agent-main` (uid 1004, `agent-*`) | `claude` (uid 1000, non-agent) |
|---|---|---|
| `gate_nonce_unit` | 19 / 0 | 19 / 0 |
| `gate_actor_path_forgery_unit` | 5 / 0 / 0 skipped | 4 / 0 / **1 skipped** |
| `gate_sudo_uid_forge_unit` | 8 / 0 | 8 / 0 |
| `gate_t2_nonce_proof_unit` | 25 / 0 | 25 / 0 |
| `gate_t2_routed_escalate_unit` | 17 / 0 | 17 / 0 |

### Correction to my own first report

I labelled those two `TIMEOUT/NO SUMMARY`. They were not slow — they exited 6 in 1s and 4s and printed no summary line, and my grep for a summary read the absence as a timeout. **A missing summary is NOT-REACHED, not slow**, and reporting it as a timeout would have buried a live regression behind a plausible-sounding one.

lodar:
## Update 2 — the migration was 3/9 done. Now complete (`b04ca2a`)

I widened my own sweep after the first two regressions and found the real blast radius: **nine harnesses stub `id()` and touch gate identity.** Removing `id -un` from the resolver strands every pin built on it. I had ported three.

### A SECOND instance of dev's defect, in `_gate_withdraw_actor`

Its non-root branch read `$EUID` and iterated `/etc/passwd` **inline**, bypassing both seams — so `gate_withdraw_unit`'s `IDUN` pin could not reach it and two arms (`Tnr2`, `T-FORGE`) graded the **runner's** identity instead of the one under test. Routed through `_gate_caller_uid` + `_gate_passwd_stream`. Semantics unchanged for the same reason as the guard: the seams' bodies *are* `$EUID` and `/etc/passwd`. `grep` now finds **zero** remaining inline bypasses in `cmd_task.sh`.

### Bridged the other six

One uniform shape that consults each harness's **own** `id()` stub dynamically, so a file re-pointing `FAKE_CALLER`/`IDUN` mid-run keeps working and no per-file pin semantics were reinvented. Production untouched: nothing in `src` reads `id`, and `src` defines both seam functions on every real invocation.

### Measured against `origin/main` at `6cd0242` — same box, same runner

| harness | main | branch (before) | branch (now) |
|---|---|---|---|
| `gate_lead_standing_unit` | 73 / 0 | **62 / 11** | 73 / 0 |
| `gate_channel_proof_unit` | 16 / 0 | **rc=6** | 16 / 0 |
| `gate_ship_routing_unit` | rc=0 | **rc=1** | rc=0 |
| `gate_withdraw_unit` | rc=0 | **rc=1** | rc=0 |
| `task_park_gate_guard_unit` | 6 / 0 | **rc=6** | 6 / 0 |
| `task_cascade_unblock_unit` | 16 / 0 | **15 / 1** | 16 / 0 |

The baseline is a real differential, not my rig.

### Final — 11 harnesses, both runners, one pinned tree

Every one `rc=0` and matching its `main` baseline on both `agent-main` (uid 1004, `agent-*`) and `claude` (uid 1000, non-agent). The only difference between runners is the newly honest `skip` on the non-agent runner.

`gate_nonce_unit` 19/0 · `gate_actor_path_forgery_unit` 5/0/0 vs 4/0/1skip · `gate_sudo_uid_forge_unit` 8/0 · `gate_t2_nonce_proof_unit` 25/0 · `gate_t2_routed_escalate_unit` 17/0 · `gate_lead_standing_unit` 73/0 · `gate_channel_proof_unit` 16/0 · `gate_ship_routing_unit` rc=0 · `gate_withdraw_unit` rc=0 · `task_park_gate_guard_unit` 6/0 · `task_cascade_unblock_unit` 16/0

### Not mine, pre-existing on `main`

`gate_tier2_decision_nonce_unit` exits 6 with no summary line on `origin/main` at `6cd0242`, unmodified tree. Reported to dev; not touched here, and I'm not claiming a cause.

lodar:
## Correction — the `gate_tier2_decision_nonce_unit` claim in this PR was too strong

I wrote that `gate_tier2_decision_nonce_unit` is "pre-existing on `main`". **It is host-specific to the control-plane box, not a `main` regression.**

dev measured both halves: the rc=6 / no-summary reproduces on a *detached, untouched* worktree at `6cd0242` (so the observation was sound and it is genuinely not mine), **but** CI run `30491990484` prints `gate_tier2_decision_nonce_unit: 12 passed, 0 failed` on that same harness. So the correct statement is: **red on this host, green in CI, cause unknown.**

The error was generalising from one environment to "`main`" without a second one — I verified the tree was untouched and never checked whether the *world* was. Nothing else in this PR depends on that claim, and the branch's own results are unaffected: they were measured on two runners on the same box, and each one matches its own `main` baseline taken the same way.

Worth someone chasing separately, for the reason I flagged earlier in this PR: a harness that exits 6 with no summary line reads as *slow* rather than *failed*.

lodar:
## Update 3 — iteration 3 was rejected on two evidence defects. One was real (`8790a3d`); one was my misreading of the reviewer's misreading.

### Real, fixed

CI on `2db4c68` finished **red**: `names_the_tree_contract_unit` reported that `tests/gate_actor_path_forgery_unit.sh` does not source `tests/lib/grading_tree.sh`, so its log named no tree. Confirmed independently before touching it — 0 `grading_tree` refs in that harness against a control of 3 in `gate_nonce_unit`.

Fixed: sourced **before** the `cd` so `${BASH_SOURCE[0]}` still resolves relative to `tests/`, and deliberately with **no** `2>/dev/null` — that redirect also swallows the helper's own stderr line, which *is* the payload, and is what once silenced all 210 harnesses at once. The harness now prints its tree and stays 5/0; the contract goes 2 passed / 0 failed.

Worth stating plainly: a harness added to close a naming-blindness defect was itself naming no tree.

### Not a defect — and I agreed with it before measuring

The reject also said `git diff origin/main...HEAD -- src/lib/registry.sh` is zero lines, and read that as the annotation removal being uncommitted. **The zero is correct and intended.** My own WIP commit `d0eeefe` *added* the mangled annotation; iteration 2 (`2802624`) removed it; so the net diff against `main` is nothing, because the thing removed was mine.

Traced per commit — `DIVE-2330` refs in `src/lib/registry.sh`:

| commit | refs |
|---|---|
| `d0eeefe` (my WIP) | **1** |
| `2802624` (iteration 2) | 0 |
| `8790a3d` (head) | 0 |
| `origin/main` | 0 |

and the mangled duplicate string occurs **0** times on the pushed object.

The companion reading — "line 41 still says *tier= is the one unforgeable field*" — is also true and also not the defect. That is `main`'s **original** sentence, which was never wrong; what I removed was my annotation glued onto it. That line being intact is evidence the removal was clean.

**My error is the instructive half.** The observation was equally consistent with an uncommitted edit and with a correctly-cancelled diff. I picked the first, invented a mechanism for it, and apologised for something that never happened — on my own branch, an hour after correcting the same shape in someone else's work. A zero diff has two causes and the sign is identical.

### The check I skipped last round and ran this time

`git show` on the **pushed object**, not the worktree: `grading_tree` refs = 2, mangled duplicate = 0, pushed head == local head `8790a3d`.

Not re-delivering until CI reaches `pending=0` on `8790a3d`. Delivering on a running check is what produced this round.

PR #332, opened 2026-07-30
task: verify --cmd must not silently replace a landed verifier result (DIVE-2067, re-land)

lodar:
**A live defect, not a missing test.** Re-land of codex's `214ac50`, graded 2026-07-26, never merged.

## What is broken on main right now

`task verify --cmd` has **no guard against closing an already-closed task**. A maker can run it over a done task and it `UPDATE`s `status`/`done_at`/`result` unconditionally, **replacing the verifier's recorded grade**.

Measured on DIVE-2059: the verifier closed with the ACK at 10:22:37; the maker closed again 39 seconds later via `verify --cmd`; two operational caveats, the red-team evidence and a follow-up split were gone. It survived only because the verifier had also compiled it to the wiki.

Reproduced on `origin/main` before writing this — the restored harness scores **2 passed / 5 failed**, with `A1 maker's verify --cmd over a done task is REFUSED` returning `rc=0`, `flippedToDone: true`, and the ACK overwritten.

## Correcting my own filing

I filed **DIVE-2389** claiming this guard was live and only its test was missing, off nine `grep DIVE-2067` hits in `cmd_task.sh`. Every one of those is a **different** guard's comment citing this incident — DIVE-2113 (start-on-closed), DIVE-2112 (reject-on-closed), DIVE-2228 (the status-writer fence). That is precisely the full-message-grep false positive DIVE-2389's own method note warns about, and I walked into it two paragraphs after writing the warning. Corrected on the row.

## The subtlety the fix preserves

`verify --cmd` is a **sanctioned escape** — DIVE-2007's refusal text names it as a real exit for a maker whose delivery was refused. So this blocks only the case with **nothing to escape from**: already done, and the closer is not the recorded verifier. A not-done task still closes normally (`B1`), and a task with no recorded verifier is untouched (`D1`).

Recommendation 3 is included: a close that still lands on a done task (the verifier re-closing their own) **appends** the prior result under a `superseded result` marker rather than dropping it.

## What I changed vs codex's commit

Content unchanged. The bundle it carried is dropped (DIVE-2091 stopped tracking it). The one addition is a **re-land note** in the code recording a real limitation:

> the actor comparison reads `task_actor`, a **provenance string a caller can set**, not the kernel-authenticated identity DIVE-2330 introduced after this fix was written.

I tried the authenticated form first — `$_svc_auth_actor` is already in scope three lines up — and **it takes C1 red**, because the harness models the verifier by setting `USER`. That is the same collision DIVE-2330 caused in nine arms of the gate suite next door. The measured incident is an accidental clobber rather than a forgery, and the two failure modes are not symmetric here: a forged actor costs one result field, whereas keying on the authenticated actor breaks every harness that models a verifier by env. So the graded form ships and the hardening (which needs a caller-uid seam in this harness) is its own row, not a re-land.

## Verification on this tree

- `tests/task_verify_over_closed_unit.sh` — **7 passed, 0 failed**
- **Mutation, both halves, reproducing codex's own controls exactly:**
  - neutralise the refusal (`if false; then`) → **5 passed, 2 failed** (A1 + A3)
  - remove the preserve/append line → **5 passed, 2 failed** (C1 + C2)
- No collateral: `task_core` 35/0, `task_verify_self_close_visibility` 10/0, `gate_lead_standing` 73/0, `task_cascade_unblock` 16/0, `task_merge_gate_ancestry` 34/0.

Part of **DIVE-2389**. Two more orphaned harnesses (`gate_tier2_nonce_evidence_unit`, `heartbeat_dam_sweep_unit`, 224 lines, from a `salvage/*` branch) are still to triage — both currently red on main and not yet diagnosed as stale-vs-real.

### DIFF
--- a/src/cmd_task.sh
+++ b/src/cmd_task.sh
@@ -3384,6 +3384,43 @@ cmd_task_verify() {
         result_txt="⚠ self-verified-close: maker=${self_verify_maker}; verifier=${self_verify_verifier} never graded; iteration=${self_verify_iteration}"$'\n'"${result_txt}"
       fi
     fi
+    # DIVE-2067: `task verify --cmd` had NO guard against closing an ALREADY-CLOSED task,
+    # so a second close REPLACED the result field outright. Measured on DIVE-2059: the
+    # verifier closed it with the ACK at 10:22:37, the MAKER closed it again 39s later via
+    # `verify --cmd`, and the ACK — two operational caveats, the red-team evidence, and a
+    # follow-up split — was silently discarded. It survived only because the verifier had
+    # also compiled it to the wiki.
+    #
+    # Note this path is a SANCTIONED escape: the DIVE-2007 refusal message names
+    # `task verify --cmd` as a real exit for a maker whose delivery was refused. So the fix
+    # must NOT close that door — it blocks only the case where there is nothing to escape
+    # FROM, i.e. the task is already done and the closer is not the recorded verifier.
+    #
+    # RE-LAND NOTE (main, DIVE-2389): this compares `task_actor`, a PROVENANCE string the
+    # caller can set, and not the kernel-authenticated identity DIVE-2330 introduced after
+    # this fix was written. That is deliberate and it is a real limitation, so read it
+    # before extending this guard. The measured incident was an ACCIDENTAL clobber (a maker
+    # re-closing 39s later), not a forgery, and the cost of the two failure modes is not
+    # symmetric here: a forged actor loses one result field, whereas keying on the
+    # authenticated actor breaks every harness that models a verifier by setting USER —
+    # exactly what DIVE-2330 did to the gate suite. I tried the authenticated form first
+    # ($_svc_auth_actor is already in scope three lines up) and it takes C1 red for that
+    # reason. Hardening it needs a caller-uid seam in this harness, which is its own row,
+    # not a re-land.
+    local _v_st _v_vfier _v_actor _v_prev
+    _v_st=$(db "SELECT COALESCE(status,'') FROM tasks WHERE id=${id};")
+    _v_vfier=$(db "SELECT COALESCE(verifier,'') FROM tasks WHERE id=${id};")
+    _v_actor=$(task_actor "")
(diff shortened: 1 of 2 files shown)


lodar:
**Correcting the verification claim in the PR body above.** It says the harness is `7 passed, 0 failed`. That was true on my box and **CI could never reproduce it** — the first run came back `6 passed, 1 failed`, plus a second red I had not run at all.

Both are re-land hazards rather than defects in the guard, and they are the same shape: *a harness restored from a stale branch has to satisfy the rules added while it was away.*

**1. `names_the_tree_contract_unit` reds on landing.** This harness was written 26 Jul, before DIVE-2211/2286 required every harness to name the tree it grades. I ran that contract against my DIVE-2383 branch and not against this one. Fixed by adding the source line, with the no-`2>/dev/null` note the contract depends on.

**2. `B1 escape preserved` failed with `rc=10 permission`, and it is not flaky.** DIVE-2015 merged as `71d5ed7` **four days after this branch was cut** and refuses the auto-close of a live delivered loop when `_gate_authenticated_actor` is empty — *"an unidentified caller cannot safely be classified as maker or verifier"*. This harness models the maker by setting `USER`/`FIVE_SENDER`, which DIVE-2330 deliberately made unforgeable. On a **non-agent runner** (CI runs as `runner`) the escape is refused before it can be exercised at all.

I am `agent-main` on the box I graded this on, so **my rig supplied the precondition** and the arm passed for a reason CI does not have. Same lesson as the one already in my notes: a baseline that shares my rig is not a baseline.

B1 is now **three-state** — it SKIPs with the reason when the caller has no kernel-authenticable identity. The skip is counted and printed separately and is **not** a pass, so a NOT-REACHED arm cannot inflate a green log.

**Proven by simulating the CI condition rather than assuming it**, since assuming is what produced the wrong number in the first place:

| condition | result | rc |
|---|---|---|
| `_gate_authenticated_actor` stubbed empty (models the CI runner) | `6 passed, 0 failed, 1 skipped` | 0 |
| real agent identity (`agent-main`) | `7 passed, 0 failed, 0 skipped` | 0 |

Everything else in the PR body stands, including both mutations — they exercise A1/A3 and C1/C2, none of which depend on the runner's identity.

PR #333, opened 2026-07-30
fix(task): name merged-PR acceptance evidence (DIVE-2217)

lodar:
Maker: **codex**. Verifier: **olivia** — this PR exists because codex has no GitHub credential; main rebased, resolved and opened it, and has **not** graded it.

## The defect

The merge-gate has three ways to accept a close, and one of them said nothing. DIVE-2184 already fixed the attribution arm's message; this is the remaining one. A close that passed on **GitHub's merged-PR record** emitted no acceptance line at all, so in the durable operator record it was indistinguishable from a close that passed on attribution. The comment directly above states the requirement it violated: *"Say WHICH evidence closed it. A close that passed on ancestry and one that passed on a merged PR are different records, and the reader of the record months later cannot tell them apart from a clean exit (DIVE-1869 rule)."*

`_merged_slug` is added for the same reason `_attr_slug` exists: the variable carrying the accepting repo is named for the arm that assigns it.

**Note what this row does NOT do**, because olivia's filing was explicit about it: `_gate_branch_ancestry` being non-accepting is **deliberate**, demoted in DIVE-2101, and filing it as a defect would point at a safety rail. That is not touched.

## Rebase, and the one real conflict

Cut from `6cd0242`, replayed onto `a6a0d98`. Both hunks conflicted with **DIVE-2266**, which landed as #327 twenty minutes ago and edits the same loop:

- `src/cmd_task.sh` — 2266 **removes** `_attr_walked` from the `local` declaration, 2217 **adds** `_merged_slug` to it. Resolved by keeping both: `_merged_slug` in, `_attr_walked` out.
- `tests/task_merge_gate_ancestry_unit.sh` — both add a line to the harness's mutation-recipe header. Resolved by keeping both entries and labelling this one `DIVE-2217:` to match the 2266 line's style.

Authorship preserved.

## Verification on the rebased tree `c07207b`

- `tests/task_merge_gate_ancestry_unit.sh` — **35 passed, 0 failed** (34 from 2266's tree, plus this row's receipt arm)
- `tests/task_merge_gate_multirepo_unit.sh` — **36 passed, 0 failed**
- **Mutation**: delete the merged-PR acceptance `warn` → **34 passed, 1 failed** at `merged-PR pass must emit its own receipt`, and the state transition still passes. That is the point — the silent-record defect reproduces while the task still closes, which is exactly why nobody noticed it.

## Not measured, and named as such by the filer

Nobody has counted how many past closes carry a misattributed acceptance line. That count decides whether this was cosmetic or a board-wide record problem, and it is not answered here.

### DIFF
--- a/src/cmd_task.sh
+++ b/src/cmd_task.sh
@@ -2105,7 +2105,7 @@ _task_status_cmd() {
         # a squash rewrites the sha, so the branch tip is NOT an ancestor of main even
         # though the content is in. Attribution finds it anyway, because the squash
         # commit subject carries the ident.
-        local _slug _bmerged="" _searched="" _attr_slug="" _anc="" _attr="" _anc_novac="" _attr_bound="" _attr_unreach=""
+        local _slug _bmerged="" _merged_slug="" _searched="" _attr_slug="" _anc="" _attr="" _anc_novac="" _attr_bound="" _attr_unreach=""
         while IFS= read -r _slug; do
           [[ -n "$_slug" ]] || continue
           _searched="${_searched:+$_searched, }$_slug"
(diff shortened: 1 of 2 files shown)


lodar:
Superseded by #342 (branch dive-2217-acceptance-arm-attribution-r2, head d6049a5), which is the same change rebased onto 0efcd66 — I diffed both patches and the only difference is a two-line comment rewrap in the test, so nothing here is lost. Closing rather than leaving it open: the DIVE-1835 done==merged gate matches the ident against open PRs and takes the first hit, so this row would have refused to close on #333 the moment #342 merged.

PR #334, opened 2026-07-30
test(gate): land the salvaged tier-2 nonce evidence harness (DIVE-2064)

lodar:
Part of **DIVE-2389**. Rescued from `salvage/untracked-test-harnesses-2026-07-26`, where it had been the only copy since 26 Jul. DIVE-2064 reads `done`; the harness never reached main.

## It scored 9/2 on main, and both failures were the codebase working

Diagnosed rather than assumed — the sibling salvage (DIVE-2067) looked stale in exactly the same way and turned out to be a **live defect** with no guard on main at all.

**1. Both audit assertions failed against a file that did not exist.** The harness predates DIVE-2010 and stubs `audit_log` alone; gate rows now route through `_task_store_audit_log`, a **fence** that calls `audit_log` only when `TASKS_DB` is the production store. On a fixture DB the row is deliberately **withheld** — so the assertions were *structurally unsatisfiable*. They failed because the fence works. Every behavioural arm (T2/T3/T4/T5/T7/T8) was already green. Fixed by opening the fence for this fixture with a `_task_human_send_allowed` stub beside the existing sink.

**2. T2 then still failed, and that one is a name move, not a fence.** It asserted the event `task answer gate`. A forged `--human` with no nonce takes the `(( human ))` arm, whose telemetry **DIVE-2233 split onto its own event**. Measured at the assertion point:

```
task answer t2-human-evidence error 0 -- task=DIVE-201 type=decision tier=2 \
  nonce_valid=0 sudo_nonagent=0 human_proof=absent
```

The row **is** written and **does** carry `tier=2` and the error verdict — everything the arm exists to demand. Only the name moved. The assertion now accepts either, since both are honest names for "a tier-2 refusal was audited".

## This corrects a claim I made earlier tonight

I wrote on DIVE-2389 that the refusal-path audit row carries no `tier=`. The `answer gate` row at ~`:7628` indeed does not — **but that is not the row this refusal writes**, and the row it does write carries `tier=` already. **No source change is needed and none is made here.** The correction is on the row.

## Verification

- **11 passed, 0 failed.**
- **Mutation:** neutralise the `task answer t2-human-evidence` audit call → T2's audit arm goes red at **10 passed, 1 failed**. Not vacuous.
- Test-only diff. No `src/` change.

## Its sibling is deliberately not here

`tests/heartbeat_dam_sweep_unit.sh` grades `_hb_dam_sweep` — a **STEER-1 function that exists nowhere on main** (`grep -rn '_hb_dam_sweep' src/` → 0, and no "dam" concept at all). Its five "passing" arms pass only because nothing runs; `command not found` on every invocation. It is a **spec, not a regression test**, and landing it would red CI.

Preserved at zero byte cost as **`refs/rescued/steer-1-dam-sweep-harness`** — the pattern dev used tonight for two commits reachable from no canonical ref — so `salvage/untracked-test-harnesses-2026-07-26` can now be retired without losing it.

### DIFF
--- a/tests/gate_tier2_nonce_evidence_unit.sh
+++ b/tests/gate_tier2_nonce_evidence_unit.sh
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+# DIVE-1448 isolated unit harness: tier-2 gate answers require CRYPTOGRAPHIC human
+# evidence regardless of need_type. Before this fix a `decision` gate floored to
+# tier 2 minted NO per-gate nonce, and the answer side TRUSTED the bare --human
+# provenance flag — which an agent forges via `sudo task answer --human`
+# (SUDO_UID=agent). Genuine tap and forged answer produced IDENTICAL rows
+# (the FUNNE-3 / DIVE-1446 incident). The fix: mint the nonce for EVERY tier>=2
+# gate, and on the answer side demand a real evidence form (valid --human-proof
+# nonce OR a non-agent SUDO_UID OR a routed lead-clear); a nonce-less LEGACY gate
+# falls back to the DIVE-1117 provenance floor for backward-compat.
+# Isolation: source src/ libs, throwaway STATE_DIR — the live shared tasks.db is
+# NEVER touched. Run: bash tests/gate_tier2_nonce_evidence_unit.sh (no root, no net).
+set -uo pipefail
+cd "$(dirname "$0")/.."
+SRC=src
+TMP="$(mktemp -d /tmp/gate-1448-unit.XXXXXX)"
+trap 'rm -rf "$TMP"' EXIT
+
+# shellcheck disable=SC1090
+for f in header.sh lib/error_codes.sh lib/output.sh lib/validation.sh \
+         lib/agent_setup.sh lib/state.sh lib/audit.sh lib/registry.sh \
+         lib/tasks_db.sh cmd_task.sh; do
+  # shellcheck source=/dev/null
+  source "$SRC/$f"
+done
+STATE_DIR="$TMP"; TASKS_DIR="$STATE_DIR/tasks"; TASKS_DB="$TASKS_DIR/tasks.db"
+GATE_PROOF_KEY="$STATE_DIR/gate-proof.key"
+GATE_PROOF_ENFORCE="$STATE_DIR/gate-proof.enforce"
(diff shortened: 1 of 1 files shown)


lodar:
Rebased onto main @6ffb695 (clean, no conflicts) and added the DIVE-2211 grading-tree block, copied verbatim from `tests/agent_git_identity_unit.sh` — including the absence of `2>/dev/null` on the source line, which the contract test calls out as load-bearing.

Measured locally at 86b2724 before pushing:

- `tests/gate_tier2_nonce_evidence_unit.sh` — 11 passed, 0 failed, **6.3s** on the (contended) control-plane host
- `tests/names_the_tree_contract_unit.sh` — 2 passed, 0 failed; corpus enumerates 269 harnesses, **0 missing** (this was the 1-failed assertion)
- `tests/meta/harness-verdict-probe.sh --only=gate_tier2_nonce_evidence_unit.sh` — **1 wired, 0 UNWIRED** (this is what the new `changed-harnesses` job runs on the PR path)
- `scripts/harness-tree-guard.sh HEAD origin/main` — rc 0

**Tier: core, by default, deliberately not demoted.** DIVE-2525 landed after this branch was cut, so the corpus is now budgeted: core is 300s/job and measured at 146s pristine / 176s installed-host. At ~4-5s in CI this is ~1.5% of the budget, so the demotion the two nearest siblings took (`gate_nonce_unit.sh` 11.0s, `gate_t2_nonce_proof_unit.sh` 10.0s, both `# TIER: nightly`) is not warranted here. Saying so explicitly because `core` is the silent default and a cost that defaults is a cost nobody reviewed.

**Overlap worth a reviewer's eye, not fixed here.** main already carries four tier-2 nonce harnesses; this is the fifth. Reading the assertion names, T2/T3/T4/T5 look substantially covered by `gate_nonce_unit.sh` (T3/T5/T6) and T1 by `gate_tier2_decision_nonce_unit.sh` (T1) and `gate_t2_nonce_proof_unit.sh` (S10). What looks genuinely new is the DIVE-1117 legacy nonce-less fallback pair (T6/T7) and the DIVE-1448 (b) tier-1 decision audit row (T8). That is the merge-by-subject case CLAUDE.md names, and it is a scope decision for this row's owner rather than something to settle while unblocking it.

PR #336, opened 2026-07-30
install: mark the merge-deploys paths at the edit site, and scope install-guard honestly (DIVE-2288)

lodar:
Maker: **main**. Verifier: **olivia** (she filed the row).

## The defect

Everywhere else in this repo, **merging is staging and the cut is the publish act.** `install.sh` is fetched from `refs/heads/main` and run **as root** on every box's next `--upgrade` — so a merge to it is live fleet-wide with no tag in between. The rule everyone was taught is *exactly backwards* in this one file, and nothing in the file, the diff, or the PR said so.

The semantics are probably right and are **not** changed here: an installer that only updates on tags cannot fix a broken installer. The defect is that the property is invisible.

## Ask 1 — visible at the edit site

- A header block in `install.sh`: **THIS FILE DEPLOYS ON MERGE. THERE IS NO TAG BETWEEN YOU AND THE FLEET.** — why the usual rule is inverted here, and where the enumeration lives.
- A `MERGE-DEPLOYS` marker at each of the **six branch-tarball fetches** (`:410` skills, `:437/:469/:497/:529/:560` 5dive-plugins). Each names what lands, that it bypasses `$REPO`'s tag pin, and the `*_TARBALL` override that *can* pin it.

## Ask 3 — CI says so, by naming what it cannot see

`install-guard.yml` opened with *"install.sh is the only file still fetched from mutable main"*. That sentence is **true and it is a sample**: scoped to files in **this repo**, while the hazard is scoped to what reaches a **box**.

**The job is not widened, because it cannot be.** A path filter is scoped to the artifact you are editing; the six other sites live in *other repos*, so no filter here could ever see them. What it can do is stop reading as exhaustive — the header now states its own scope, and the summary tables the six paths it is blind to. A guard that names its blind spot beats one that reads as complete.

A real cross-repo guard would have to live in `5dive-ai/skills` and `5dive-ai/5dive-plugins`. Not built here, and worth its own row if we want it.

## Ask 2

Already done as **DIVE-2308**.

## Correction to the enumeration — measured, not relayed

DIVE-2308 read `origin/main @ 44e612f` on 07-29 and listed 9 sites. Re-measured on **`6f5e5c9`**: **row 8 is already fixed.** `_published_cli_probe` no longer does `ref="${pinned:-main}"` — it resolves the newest release tag through three rungs (`ls-remote` → `tags.atom` → tags API) and **fails closed** with *"no release tag resolves"* when none does. Main HEAD is never called "published" any more.

So the live **install-path** population is **7, not 9**: row 8 closed, and row 9 (the `status` badge branch) is public-by-design and off the install path. The guard text says 7 for that reason, not because it disagrees with DIVE-2308.

## Verification

- `shellcheck -S error install.sh` — clean
- YAML parses; `build.sh` builds
- `install_checksum` 8/0 · `install_monotonicity` 14/0 · `install_pin_sha` 15/0 · `install_update_hint` 6/0 · `update_check_propagation` 22/0
- `install.sh --help` returns `rc=1` **before and after** — pre-existing, the installer has no `--help`. Checked against the base rather than assumed.

**Comments and CI prose only. No executable line of `install.sh` changed.**

## What I did not check

The contents of `5dive-ai/skills` and `5dive-ai/5dive-plugins` themselves — only the fetch URLs in this repo that target their mutable `main`. And I did not touch the `status` branch, which is merge-deploys by design.

---

## Iteration 2 (`17f7998`) — read this before the diff

olivia rejected iteration 1 on one thing: the commit published **two populations** (header said 9, CI said 7) and left the authoritative wiki page stale, with the correction living only in the commit message and the task result. That is this row's own defect, committed by the change meant to fix it.

Fixed: the count now carries **its scope and its date** in both artifacts *in the same words* — "7 live sites on the root-installed-on-upgrade rail, as of 2026-07-30, measured against `6f5e5c9`" — and the wiki page gained a **currency marker**, row 8 struck as **CLOSED** with evidence, and `corrected_against: 6f5e5c9`. `_marketplace_base` and `_loops_base` are named in all three artifacts as adjacent to the count and deliberately outside it (on-demand, not root-installed on `--upgrade`; verified neither is referenced in `install.sh`).

Full detail in the iteration-2 comment below.

### DIFF
--- a/.github/workflows/install-guard.yml
+++ b/.github/workflows/install-guard.yml
@@ -1,12 +1,23 @@
 # DIVE-2144, scope item 4: put the root-path diff in front of a human on every PR.
 #
-# After the tag cutover, install.sh is the only file still fetched from mutable main
-# and executed AS ROOT on every box (cmd_selfupdate.sh re-fetches and runs it on each
-# upgrade). A change to it is live on the whole fleet the moment it merges, with no
-# tag in between. This job does not block that — CODEOWNERS plus branch protection is
-# what blocks it. This job makes it IMPOSSIBLE TO MERGE ONE WITHOUT HAVING SEEN IT:
-# the full diff lands in the job summary, so "I didn't realise that PR touched the
-# root path" stops being available as an explanation.
+# After the tag cutover, install.sh is the only file IN THIS REPO still fetched from
+# mutable main and executed AS ROOT on every box (cmd_selfupdate.sh re-fetches and runs
+# it on each upgrade). A change to it is live on the whole fleet the moment it merges,
+# with no tag in between. This job does not block that — CODEOWNERS plus branch
+# protection is what blocks it. This job makes it IMPOSSIBLE TO MERGE ONE WITHOUT
+# HAVING SEEN IT: the full diff lands in the job summary, so "I didn't realise that PR
+# touched the root path" stops being available as an explanation.
+#
+# DIVE-2288/DIVE-2308 — READ THE SCOPE OF THE SENTENCE ABOVE. "The only file in this
+# repo" is TRUE and it is a SAMPLE, not the population. The risk this job exists to
+# bound is what reaches a customer BOX without a tag, and by that boundary install.sh
+# is 1 of 7 live merge-deploys sites. The other six are BRANCH TARBALLS of
+# 5dive-ai/skills and 5dive-ai/5dive-plugins, fetched by install.sh and root-installed
+# on every --upgrade. NO PATH FILTER IN THIS REPO CAN EVER SEE THEM, because a path
+# filter is scoped to the artifact you are editing while the hazard is scoped to where
+# the bytes land. So this job cannot be widened to cover them — it can only SAY SO,
+# which the summary below now does. A guard that names its own blind spot is worth more
+# than one that reads as exhaustive.
 name: install-guard
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


lodar:
**Iteration 2** — `17f7998`. olivia rejected iteration 1 on one thing and was right.

## The reject

The commit published **two populations** and left the authoritative one stale. `install.sh`'s header said *"9 sites across 4 repos"* and pointed at the wiki page; `install-guard.yml` **in the same commit** said *"1 of 7"* twice; and my own measurement said 9 was stale because row 8 is closed. The correction lived only in the commit message and the task result — **the two places a reader never looks.**

That is the exact defect this row exists to kill, committed by the change meant to fix it: *a true-when-written figure presented at an edit site with no currency marker.*

## Fixed in one pass

**(a) The count now carries its scope and its date, in both artifacts, in the same words** — *"7 live sites on the root-installed-on-upgrade rail, as of 2026-07-30, measured against `6f5e5c9`"*, plus what changed from the 9 (row 8 closed, row 9 by design). `install.sh` and `install-guard.yml` can no longer disagree.

**(b) The wiki page carries the correction**, since `install.sh` sends people there:
- a **CURRENCY MARKER** at the top — read before quoting a number from this page
- row 8 struck and marked **CLOSED**, with the evidence (three-rung tag resolution, fails closed, no `pinned:-main` in the tree)
- total restated; `corrected: 2026-07-30` / `corrected_against: 6f5e5c9` added to frontmatter
- the old row 8 text **kept** under *"What was true on 44e612f"* — the shape it names is the durable lesson even though the state it reports is not current
- index line updated

## Her non-blocking call, answered by scoping rather than renumbering

`_marketplace_base` (`src/cmd_pack.sh:28`) and `_loops_base` (`src/cmd_loop_pack.sh:16`) do each fetch another repo's mutable main onto a box. **Verified on `6f5e5c9` that neither is referenced anywhere in `install.sh`** — they run **on demand** when a user invokes the verb, not root-installed on every `--upgrade`.

Different rail. Folding them into the 7 conflates two things; leaving them unnamed repeats the sample-vs-population defect. So both are now named explicitly — in the header, in the CI header, and on the wiki — as **adjacent and deliberately outside the count**.

## Re-verified

`install_checksum` 8/0 · `install_monotonicity` 14/0 · `install_pin_sha` 15/0 · `install_update_hint` 6/0 · `update_check_propagation` 22/0. `shellcheck -S error` clean, YAML parses.

Consistency **checked rather than assumed**: zero occurrences of the old live *"9 sites across 4 repos"* figure remain in `install.sh`; `install-guard` says *"1 of 7"* twice, both scoped and dated.

Still comments and CI prose only. No executable line changed.

PR #341, opened 2026-07-30
fix(tasks-db): remove five printf-into-grep SIGPIPE sites in schema migration (DIVE-2434)

lodar:
Unblocks the release cut: this is the single red check on main's tip `0efcd66`.

## What actually happened, corrected

`src/lib/tasks_db.sh:1007` ran `printf '%s\n' "$cols" | grep -qx "${c%% *}"`. grep can exit on its first match while printf is still writing; printf takes SIGPIPE and emits `write error: Broken pipe` on stderr.

**This is product code, not test code.** It sits in `_tasks_db_migrate`, called from `tasks_db_init` on essentially every `5dive task` invocation — so the stray line can surface anywhere, and it will keep recurring intermittently while looking like flake. (My first report attributed it to the harness. olivia corrected that, and the correction is what makes it worth fixing rather than re-running.)

## How a cosmetic stderr line turned a green test red

`tests/loop_panel_unit.sh` T6 runs the subject as `( cmd_loop_panel … >"$TMP"/panel-pass.out 2>&1 ) &` and then does `jq -r '.data.verdict' "$TMP"/panel-pass.out`. The stderr line landed **inside the captured stream the assertion parses**. jq got a file whose first line is not JSON, `pv` came back empty, and the assertion failed — while the loop panel itself had succeeded.

That is visible in the CI log and I misread it: the failure message is `bad_t "quorum pass" "$(cat …/panel-pass.out)"`, so the broken-pipe line printed immediately above a healthy `status:done, haltReason:complete` **is the captured file**, not two independent events. olivia proposed this reading as an alternative to mine; confirmed at `tests/loop_panel_unit.sh:132/142/145`. **Nothing about quorum was broken.**

## The mechanism is NOT established, and the fix does not depend on it

olivia built the exact shape at the real payload size (~70 short strings, far under the 64KB pipe buffer) with the match on line one, ran it 400 times, and got **zero** broken pipes. So the obvious early-exit story does not reproduce where it actually occurs. CI under load is the untested variable.

Rather than ship a diagnosis that failed its own check, this removes the pipe. No pipe, no SIGPIPE, whichever race fires. Class removal, not repair.

## Verification

- Equivalence checked **independently on both sides**, not relayed: exact match, no match, and the prefix case `grep -x` must reject (`need_ty` vs `need_type`) return identical verdicts pipe and herestring.
- `tests/loop_panel_unit.sh`: **14 passed, 0 failed** on this branch.
- Branched from `0efcd66` so CI here grades the same tree that went red.

## Latent hazard, named at the site

If anyone adds `set -o pipefail` on this path, the SIGPIPE stops being cosmetic and becomes a hard failure in schema migration.

## Note for the release

Independent of this fix: `release-cut` refuses on **any** non-success check-run, so the cut is gated on total repo health rather than release readiness. Three nights of `assign`+`version-uniqueness`, then this the moment those were deleted. That is a design property and it will keep producing brakes nobody chose.

### DIFF
--- a/src/lib/tasks_db.sh
+++ b/src/lib/tasks_db.sh
@@ -1292,7 +1292,7 @@ MIG
     local lr_cols
     lr_cols=$(sqlite3 -cmd ".timeout 5000" "$TASKS_DB" \
               "SELECT name FROM pragma_table_info('loop_runs');" 2>/dev/null)
-    if ! printf '%s\n' "$lr_cols" | grep -qx "scorecard_json"; then
+    if ! grep -qx "scorecard_json" <<<"$lr_cols"; then
       sqlite3 -cmd ".timeout 5000" "$TASKS_DB" \
         "ALTER TABLE loop_runs ADD COLUMN scorecard_json TEXT;" >/dev/null 2>&1 || true
     fi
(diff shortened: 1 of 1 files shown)


lodar:
Re-identified onto **DIVE-2434** (its own row). This PR was opened 05:52, 26 min before the DIVE-2418 row existed, and the v0.17.9 cut completed at 06:18 without it, so it was never DIVE-2418's deliverable — it only carried that ident because the branch was cut while chasing the 0efcd66 red. olivia overrode the DIVE-1835 merge gate on closing 2418 (audited) and asked for the re-ident so this does not go orphaned. DIVE-2247 was the other candidate and closed at 06:18:43, so it would have bound a live PR to a closed row.

PR #348, opened 2026-07-30
fix(gate): the done==merged gate searches every shipping repo, and the ACCEPT names its scope (DIVE-2431)

lodar:
`_gate_repo_slugs` defaulted to three slugs — `5dive-ai/5dive`, `lodar/5dive-api`, `lodar/5dive-frontend` — so a delivery to any other repo we ship from was graded by a set that never contained it.

Both failure directions were live, measured on DIVE-2303 whose delivery landed in `5dive-ai/character-packs`:

- **False accept.** It closed clean on a `5dive-ai/5dive` commit from step 1 of the same ticket, having never looked at character-packs. Correct by luck.
- **False refuse.** Strip that coincidence and genuinely-landed work is refused, with an audited `--force-merge-gate` as the remedy for a gate that was simply looking in the wrong place.

## What changed

**Widened to the 11 active product repos.** A list and not a derivation: deriving the set from the box's git remotes or the org API removes the drift but makes the verdict depend on host filesystem state or network reachability at close time. A gate that answers differently on two boxes, or refuses when offline, is not a gate.

**Every verdict now names the set it searched.** The refusals always did; the ACCEPT did not, and that was the silent half — an accept sourced from a repo that is not where the delivery went reads as a clean close and nothing invites a second look. Only stated when the task declares no repo, since a declared repo narrows the scan to itself and leaves no unsearched remainder. That disclosure is what pays for the drift in the hardcoded list: a stale set now announces itself at the moment it matters, to the person it is failing.

## Tests

`tests/task_merge_gate_repo_scope_unit.sh` — new, 25 assertions, graded on both properties independently so a future widening that drops the disclosure goes red.

The default-set canary in `task_merge_gate_inferred_repo_unit.sh` went red the moment the list moved, which is exactly its job. Updated rather than relaxed — the superset precondition it guards is stronger with a bigger set.

Local: 25/0, 18/0, 35/0 across the three gate suites; shellcheck clean at `-S warning`.

### DIFF
--- a/src/cmd_task.sh
+++ b/src/cmd_task.sh
@@ -1453,10 +1453,38 @@ _gate_text_names_a_ref() {
 # coverage AND — worse — an api task naming "PR #6" got a CONFIDENT verdict about an
 # unrelated CLI pull request. Overridable so a new repo is config, not a patch, and
 # so the tests can point the whole gate at fixtures.
+# DIVE-2431: the default was exactly those three, and a delivery to any OTHER repo we
+# ship from was graded by a set that never contained it. Both directions were live and
+# both were measured on DIVE-2303, whose delivery landed in 5dive-ai/character-packs:
+#   FALSE ACCEPT  — the gate found the ident in 5dive-ai/5dive (step 1 of the same
+#                   ticket, days earlier) and closed clean having never looked at
+#                   character-packs. Correct by luck.
+#   FALSE REFUSE  — strip that coincidence and genuinely-landed work is refused with
+#                   "nothing on main in <3 slugs> shows branch ... landed", whose remedy
+#                   is an audited `--force-merge-gate` override for a gate that was
+#                   simply looking in the wrong place.
+#
+# WHY A LIST AND NOT A DERIVATION. Deriving the set from the git remotes on the box, or
+# from the org's repo list, removes the drift but buys a worse property: the gate's
+# verdict would then depend on host filesystem state or on network reachability at close
+# time. A gate that answers differently on two boxes, or refuses when offline, is not a
+# gate. The list is deterministic and auditable; drift is the price.
+#
+# WHAT PAYS FOR THE DRIFT: every verdict names the set it searched — the refusals always
+# did, and DIVE-2431 added it to the ACCEPT, which was the silent half. A stale list now
+# announces itself at the exact moment it matters, to the person it is failing. That is
+# the property to preserve if this list is ever edited; adding a repo without it just
+# moves the blind spot.
 _gate_repo_slugs() {
   local raw="${FIVE_GATE_REPOS:-}"
   if [[ -z "$raw" ]]; then
+    # The repos task deliveries actually land in. Kept to ACTIVE product repos rather
+    # than every repo we own — an inactive repo costs a lookup on every close and has
+    # never received a delivery.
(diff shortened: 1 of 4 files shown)


lodar:
## Measured, so the list is not a judgement call

**Every added repo genuinely receives deliveries.** DIVE-subject commits in each repo's last 100 on main:

| repo | DIVE-subject commits |
|---|---|
| 5dive-ai/5dive-plugins | 82 |
| 5dive-ai/openagent | 49 |
| 5dive-ai/character-packs | 37 |
| 5dive-ai/skills | 27 |
| lodar/5dive-mobile | 20 |
| 5dive-ai/ops | 15 |
| 5dive-ai/5dive-mcp | 2 |
| lodar/5dive-blog | 2 |

So all eight were live delivery targets the gate could not see, not repos added because we happen to own them. The two 2s are thin and are the first candidates to drop if the cost ever matters.

## The cost, stated rather than left unmeasured

One commit-list call is **0.6–1.0s** per repo, so 3 → 11 adds roughly **5–8s** — but only on the path that actually walks the whole set.

Two things bound it: the scan **breaks on the first hit**, and `5dive-ai/5dive` is **first** in the list (pinned by an assertion in the new suite). A close whose work is in the CLI repo still costs one lookup. Only a close whose delivery is elsewhere — exactly the case that was previously answered wrongly — pays the walk, and it pays it to get a correct answer instead of a fast wrong one.

A task that declares `Repo:` or binds a delivery_ref narrows the scan to one repo and is unaffected.

lodar:
## The risk this widening carries, surfaced by CI rather than by my design

CI went red on `T6` in `task_merge_gate_diagnostic_unit.sh`, and the cause is worth reading as a property, not just a fixture gap.

**T6 asserts that a genuine miss refuses under DIVE-1830.** That only holds if *every* repo in the search set answers — an unanswered repo makes it DIVE-2318's `PARTIAL COVERAGE` refusal instead. The fixture stubbed the three slugs the default happened to contain, so eight of eleven read as unreachable and the arm went red for a reason unrelated to its subject.

**The generalisation: with 11 repos instead of 3, any repo a caller's token cannot read degrades that close to "partial coverage, inconclusive".** That blast radius is wider than it was, and it is the real cost of this change.

Three things bound it, and I think they make it acceptable:

1. **Measured:** all 11 are readable by the standard agent token today (checked individually, not assumed).
2. **The failure is honest.** `done-attribution-unresolved` states it is partial coverage, names the repo it could not read, and gives the remedy ("hand the close to an agent whose token can see it", or narrow with a `Repo:` line). It is a refusal with a named next step, not a wrong verdict — which is strictly the direction this ticket exists to move things.
3. **`FIVE_GATE_REPOS` still overrides**, so a box with a narrower token can narrow the set.

If it does bite, the two thinnest entries by delivery history are `5dive-ai/5dive-mcp` and `lodar/5dive-blog` (2 DIVE-subject commits each in their last 100) and they are the first to drop.

**Fixture fix, and the part worth keeping:** T6 now stubs whatever `_gate_repo_slugs` returns rather than naming the set itself, and refuses to run against a set smaller than 3 rather than passing for the wrong reason. A fixture that names a set the code owns is a second copy of that set, and the copy is the one nobody updates.

All 15 gate-touching suites — enumerated by `grep -rl`, after I initially ran three from memory and CI caught the one I missed — are green and match a clean `git archive` baseline of `origin/main`.

PR #367, opened 2026-08-02
feat(actor): 5dive whoami — one sealed uid-first actor derivation that refuses when unmeasurable (DIVE-2517)

lodar:
## DIVE-2517 — v0.18 W1: one sealed actor derivation

The CLI had **six** actor derivations and they disagreed. Only `_gate_authenticated_actor` failed closed. The rest resolved identity from something the caller can set — `--from` on argv, `$USER`, `$SUDO_USER`, `FIVEDIVE_AUDIT_USER`, or an `agent-` username prefix — and none refused when they could not measure.

`src/lib/actor.sh` promotes the uid-first resolver into the single derivation, and `5dive whoami` reports what it measured **and where each value came from**.

### What it derives

| axis | source | forgeable by the caller? |
|---|---|---|
| identity | `$EUID` (kernel builtin) → `/etc/passwd`, pure bash | no |
| agent name | the **registry's** answer, not a username prefix | no |
| authority | `_actor_authority` in `lib/audit.sh`, **wired not rewritten** | no |
| tier | `registry .agents[<agent>].isolation` | no |

No `id`, no `getent` — both resolve through the caller's `PATH` (DIVE-2330). `$SUDO_UID` is honoured only at real EUID 0, where forging it would require already being root.

**Unmeasurable is `exit 6`**, not the word `unknown` printed with `rc=0`. That refusal is the point of the verb.

### The regression this PR also fixes — and how it was caught

Promoting six functions out of `cmd_task.sh` broke every harness that names its source files explicitly. 89 lists jumped straight from `lib/tasks_db.sh` to `cmd_task.sh`, modelling a bundle that **cannot exist** — `build.sh` puts `actor.sh` between them.

Baselined at `51bc5c6` before believing it. All five were 0-failed on main; **26 arms** red on the feature commit:

| harness | main | feature commit | now |
|---|---|---|---|
| `gate_lead_standing_unit` | 73/0 | 62/**11** | 73/0 |
| `gate_t2_routed_escalate_unit` | 17/0 | 10/**7** | 17/0 |
| `task_verify_self_close_visibility_unit` | 10/0 | 3/**7** | 10/0 |
| `gate_nonce_unit` | 19/0 | **refused at the seam pin** | 19/0 |
| `push_unit` | 82/0 | 81/**1** | 82/0 |

Worth naming *which* one failed usefully: `gate_nonce_unit` pins its seam before any arm runs — it asserts `_gate_uid_to_agent` resolves to `evil` and aborts if not — so a resolver that had vanished from the sourced set **refused the run** instead of grading an absent function. The other four had no such pin and quietly went red. That pin is the reason this was one grep rather than an afternoon.

`push_unit:346` needed a *different* fix, not the same one. It unsets `_gate_agent_for_uid` and re-sources to restore it; that helper moved too, so restoring from `cmd_task.sh` would have left it **undefined** and the arms below would have graded an absent function — green for the wrong reason. It now sources `lib/actor.sh`.

### How it was checked

`tests/whoami_sealed_actor_unit.sh` — **18 passed, 0 failed**. Mutation-graded three ways, each caught by its intended arm rather than by the harness name:

| mutant | caught by |
|---|---|
| unmeasurable exits `0` instead of `6` | T12 — *"the bundle exited 0 with an unmeasurable actor"* |
| identity consults `$SUDO_USER` | T3 — *"env forgery moved the identity to 'agent-olivia'"* |
| residual `unknown:*` `continue`s instead of returning | T17 — *"papered over by candidate 2"* |

T13/T14/T18 fail explicitly on **collapse** and on **vacuity** — they refuse when two registry verdicts render identically, or when both names resolve the same way and the registry is demonstrably not deciding.

The seam contract in the ticket is kept: `$EUID` is read at exactly two places, `_gate_is_root` and `_gate_caller_uid` (`actor.sh:91-92`). No second seam. `lib/audit.sh` is untouched.

### Honest limits

- The **local 264-harness sweep was still running** when this PR was opened (51/264, 0 failures at that point). CI is the authority on the full corpus; if it disagrees with my partial sweep, believe CI.
- `--from` is rejected as an unknown flag (`rc=2`), so the verb takes no identity from argv. The *other five* derivations still exist and still disagree — this PR builds the sealed one and wires the gate path to it. Retiring the remaining callers is not in this row.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## Unreleased — feat(actor): `5dive whoami`, one sealed actor derivation (DIVE-2517)
+
+The CLI had **six** actor derivations and they disagreed. Only one failed closed;
+the rest resolved identity from something the caller can set — `--from` on argv,
+`$USER`, `$SUDO_USER`, `FIVEDIVE_AUDIT_USER`, or an `agent-` username prefix —
+and none of them refused when they could not measure.
+
+`src/lib/actor.sh` promotes the uid-first resolver into the single derivation.
+Identity comes from `$EUID`, a kernel-backed bash builtin, mapped through
+`/etc/passwd` in pure bash — no `id`, no `getent`, both of which resolve through
+the caller's `PATH` (DIVE-2330). `$SUDO_UID` is honoured only at real EUID 0,
+where forging it would require already being root. Agent-ness is the registry's
+answer, never a username prefix. Authority is not redefined here: `_actor_authority`
+in `lib/audit.sh` already single-sources `root` / `sudo:<who>` / `self`, and the
+verb wires it.
+
+`5dive whoami` reports actor, authority and tier **with the source of each**, plus
+`--json`. An actor it cannot measure is an **exit 6**, not the word `unknown`
+printed with `rc=0` — that refusal is the point of the verb, and it is
+mutation-graded rather than asserted.
+
 ## Unreleased — fix(task): refuse a close that would REPLACE an already-closed row's result (DIVE-2464)
 
 `5dive task done <id> --result=...` on a row that was already done overwrote the result column and
--- a/build.sh
+++ b/build.sh
@@ -36,6 +36,7 @@ cat \
   src/lib/audit.sh \
   src/lib/registry.sh \
   src/lib/tasks_db.sh \
+  src/lib/actor.sh \
   src/cmd_auth.sh \
   src/cmd_account.sh \
   src/cmd_agent.sh \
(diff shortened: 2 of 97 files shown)


lodar:
Superseded by a fresh PR from the same branch to get CI. This PR fired only bundle-drift: the pull_request event is suppressed for it, and BOTH documented remedies failed on measurement — close/reopen produced zero pull_request runs, and an empty commit pushed with a different (OAuth) credential produced zero as well, while an ordinary branch the same morning got unit-tests + pii-guard + install-smoke + supply-chain-guard. So the suppression follows the PR, not the event actor. Reopening from scratch is the only lever left. Code is unchanged and byte-identical.

PR #372, opened 2026-08-02
revert(release): restore the last-parseable release-cut.yml (DIVE-2539)

lodar:
My DIVE-2539 change made release-cut.yml invalid to GitHub Actions — zero-job 0s failures on every commit since 15:25, and the 02:37Z nightly cut would not fire. Restoring the last-parseable bytes; the level input gets redone separately once I can validate it.

`yaml.safe_load` accepting the file is not validation — Actions applies a stricter schema. The available tell was `gh workflow list` showing the raw path instead of `release-cut`, which is gh's fallback when the name cannot be parsed.

### DIFF
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -69,15 +69,6 @@ on:
         description: 'Why this cut is being made by hand (hotfix, incident, first cut)'
         required: true
         type: string
-      level:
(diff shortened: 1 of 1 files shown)


lodar:
Superseded by the expression fix — actionlint identified the real cause (an empty Actions expression delimiter inside a shell comment), so main can go straight to a working cutter WITH the level input rather than reverting to patch-only.

PR #399, opened 2026-08-03
SECURITY(gate): an agent could forge a human tap on a tier-2 decision gate with one env var (DIVE-2588)

lodar:
GATE_PROOF_ENFORCE=/nonexistent/nope  5dive task answer <T2-decision> --value=A --human

succeeded, unprivileged and without sudo, where the identical command without
the variable was refused — and wrote need_answered_by=human:<agent>, the same
row the human's own relayed tap writes. Found by main on the released v0.18.2
while grading DIVE-2515; reproduced here on the installed 0.18.0+dive2563 as
agent-dev uid 1007 in an isolated store, baseline and bypass as a pair.

_gate_proof_enforced was a file-EXISTENCE test on a path the constrained party
supplies. Its OFF position was reachable by its own subject, which is not a
control. Scope was exactly 'decision': approval/secret/manual/access carry an
independent type guard that never consults the flag.

Two independent halves, because either alone closes the reported command while
leaving the shape intact:

1. src/lib/tasks_db.sh — the override may only make enforcement STRICTER. A
   sentinel existing at EITHER the root-owned default or $GATE_PROOF_ENFORCE
   arms it; an absent path now says nothing instead of saying 'off'. Every
   caller in tests/ points the override at $STATE_DIR/gate-proof.enforce, the
   default it would have resolved anyway, so the harness pattern is untouched.
2. src/cmd_task.sh — the tier-2 provenance floor and the DIVE-2356 evidence
   block no longer consult the flag at all. It was a rollout envelope for a
   rollout that completed 2026-07-30, and while it stood it made a hard human
   floor switchable. The floor refuses only a NON-human answer, and the
   evidence block is already scoped to gates carrying a minted nonce, so a box
   that mints nothing reaches no new assertion (DIVE-525 holds).

Also: 'gate-proof enforce off' no longer leaves the default sentinel behind
(it would have printed OFF while the predicate read ON), and 'status' names
which sentinel armed it.

selfcheck: new t2-forge probe performs the forge for real in a throwaway store
and reds if it lands — DIVE-2520's forge-an-actor probe one layer up. It also
clears a second gate down a real human path, because a floor that refuses
everything and a floor that refuses the forge are the same observation from the
refusal alone. Runs on every PR in both unit-tests.yml selfcheck jobs. Its
first run delivered a real inter-agent message about a fixture gate: the answer
path's resume ping rides cmd_send, which no STATE_DIR fence covers, so that
rail is stubbed in the probe and in the new harness.

tests/gate_enforce_env_bypass_unit.sh (19 arms, TIER: nightly — the core tier
already reads 325s/300s). Mutation matrix, by failing assertion NAME:
  pristine                      19/0
  predicate reverted            18/1  P2
  floor conjunct restored       18/1  E5
  evidence conjunct restored    18/1  E6
  all three (the pre-fix code)  13/6  P2 E2 E3 E4 E5 E6, with the exploit
                                      reproduced inside the harness as
                                      state=closed prov=human:dev
tests/gate_tier2_floor_unit.sh and tests/gate_t2_routed_escalate_unit.sh each
carried an arm asserting the OPPOSITE contract — that clearing the flag made
the floor dormant — and now assert that it does not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,57 @@
 # Changelog
 
+## Unreleased — SECURITY fix(gate): a tier-2 human floor whose OFF switch was reachable by the agents it constrains (DIVE-2588)
+
+Any agent, unprivileged and without sudo, could forge a human tap on a tier-2 **decision**
+gate with one environment variable:
+
+```
+GATE_PROOF_ENFORCE=/nonexistent/nope  5dive task answer <T2-decision> --value=A --human
+```
+
+succeeded where the identical command without it was refused, and wrote
+`need_answered_by=human:<agent>` — the same row the human's own relayed tap writes, so the
+record could not tell them apart. Found by main while grading DIVE-2515 on the released
+v0.18.2 artifact; reproduced on the installed 0.18.0+dive2563 in an isolated store.
+`_gate_proof_enforced` was a file-**existence** test on a path the constrained party
+supplies, so pointing it at anything absent evaluated the whole tier-2 branch false. Scope
+was exactly `decision`: approval/secret/manual/access carry an independent type guard that
+never consults the flag.
+
+Two independent halves, because either alone would have closed the reported command while
+leaving the shape intact:
+
+- **The override may only make enforcement stricter.** An existing sentinel at *either* the
+  root-owned default or the `GATE_PROOF_ENFORCE` path arms it; an absent path now says
+  nothing at all instead of saying "off". Harnesses keep the override they already use —
+  every caller in `tests/` points it at `$STATE_DIR/gate-proof.enforce`, the default it
+  would have resolved anyway.
+- **The tier-2 human floor no longer consults the flag**, in either direction. It was a
+  rollout envelope for a rollout that finished 2026-07-30, and while it stood it made a hard
+  human floor switchable. Safe to make unconditional: the provenance floor refuses only a
+  *non-human* answer on a tier-2 gate, and the evidence block is already scoped to gates
+  carrying a minted nonce — so a box that mints nothing reaches no new assertion, and every
+  real human path passes `--human` (DIVE-525 holds by construction).
+
+Behaviour change worth naming, and its bound: on a box where the sentinel was never armed,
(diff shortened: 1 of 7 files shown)


lodar:
**Update — `4fbd7c8`, 13 checks green.** The first push (`9e0f16f`) was red in three jobs and the cause was mine, not the environment.

**The harness and the `t2-forge` probe were grading the host, not the fix.** Both stubbed `id -un` to model an agent caller. The uid-first resolver stopped reading `id -un` at DIVE-2330 — it walks `/etc/passwd` for `$EUID` in pure bash, precisely so a PATH shim cannot forge it — so the stub was **inert** and agent-ness came from uid 1007 being `agent-dev` on a 5dive box. 19/19 green graded nothing. On a runner that uid is `runner`: no agent claims it, `--human` is a **legitimate human tap**, the "forge" arm was never a forge, and 8 arms went red starting with the baseline refusal the file rests on.

Reproduced locally by uid alone — `sudo -u claude` (a non-agent uid), same box, same store — pre-fix reds with CI's exact eight arm names, post-fix passes 23/23.

Both sites now pin `_gate_caller_uid` + `_gate_passwd_stream` (the seams `lib/actor.sh` publishes for this) and assert the pin **through the real resolver** before any arm depends on it. The probe reports `not-reached`, never `pass`, when its caller does not resolve as an agent.

Also in this commit: `seed_task` reds on a duplicate ident instead of emitting a bare `UNIQUE constraint failed` (DIVE-406 was seeded twice — harmless to the verdict, but it printed one line above the failures and cost a reviewer the root cause); and I1/I2 prove store isolation **before the first write, differentially** per DIVE-2518.

Mutation matrix re-taken at **23 arms** (was 19), same failing assertion names: pristine `23/0` · predicate reverted `22/1` (P2) · floor conjunct restored `22/1` (E5) · evidence conjunct restored `22/1` (E6) · all three, the pre-fix code, `17/6` (P2 E2 E3 E4 E5 E6) with the exploit reproduced in-harness at `state=closed prov=human:dev`. Header timing corrected 8.6s → 9.5s.

The security change itself is unchanged from `9e0f16f`.

PR #405, opened 2026-08-03
fix(pack): the skills text told codex/opencode users NONE are installed while import lands bodies (DIVE-2583)

lodar:
## The contradiction

Every exported `AGENTS.md` said:

> They travel as NAMES, not bodies … On a harness **with no skills directory (codex, opencode) NONE are installed**.

Both claims are wrong, and the code that contradicts them is two files away:

- `SKILLS_INSTALL_DIR` maps **codex → `.agents/skills`** and **opencode → `.agents/skills`**. An *unmapped* type does not get nothing either — it falls back to `.claude/skills`. The category "harness with no skills directory" is **empty**.
- `_install_bundled_skill` is **not type-gated**: it resolves that map for any type and `cp -r`s the body. `cmd_skill_add` (the fallback path a single-file import actually takes, since AGENTS.md carries refs) resolves the same map. So importing onto a codex seat **does** land bodies at `/home/agent-<n>/.agents/skills/<id>/` and lists them in `skillsAdded` — while the file the user is reading says none are installed. That is the permissive direction.

## Why it survived review

The two statements are about different **directions**, and one sentence fused them:

| direction | true statement |
|---|---|
| export (this file) | carries skill NAMES, never bodies — for **every** harness |
| import | resolves each name and installs bodies to `skills_install_dir(type)` — for **every** harness |

Each half is defensible alone. The falsehood lives only in the joint: a premise scoped to export welded to a parenthetical about the harness, which reads as a universal fact about codex.

## The fix

The sentence was a **fork of the map** — someone read `SKILLS_INSTALL_DIR` once, restated it in prose, and the copy aged independently. Better wording does not fix that, so the sentence now has no independent opinion:

```sh
skills_install_dir() { printf '%s\n' "${SKILLS_INSTALL_DIR[${1:-}]:-.claude/skills}"; }
```

called by the installer **and** by the renderer, which prints the resolved dir for the pack's own type. Claim and behaviour read the same array.

## Left unmeasured, on the record

- **Whether codex/opencode ever SCAN `~/.agents/skills`** — no codex seat on this box (its installer fails partway). The render states the *mechanism* ("a body lands in `~/.agents/skills/<skill>/`; whether codex then LOADS that directory is the harness's own behaviour and 5dive does not verify it") and never the outcome. Recorded at the map entry too.
- **The `antigravity` `{workspace}`-vs-`$HOME` discrepancy** folded in as asked: its own comment says agy reads `{workspace}/.agents/skills`, while every writer we have joins the value to `$HOME` — and an agy seat's workspace is a project dir. If the comment is right those bodies are present-and-inert. No antigravity agent exists in the fleet (DIVE-2037), so it is recorded, not resolved.

## Grading — the predicate, not the string

15 arms in `tests/pack_agents_md_unit.sh` (52 → 67, 1.0s, core). The acceptance criterion asked for an arm that **executes the predicate** rather than grepping the warn string. The load-bearing one renders once per mapped type and requires the dir in the prose to **equal `skills_install_dir(type)`** — a hardcoded string passes for at most one type out of eight.

Mutation-graded (5 mutants against the committed tree, all restored):

| mutant | reds |
|---|---|
| hardcode the rendered dir | `rendered dir == resolver`, `codex export names .agents/skills` |
| `[codex]=""` in the map | map arm, resolver-agreement arm, codex arm, codex render arm |
| resolver returns a constant | resolver-agreement, codex, opencode, codex render |
| old false sentence back in the render | both negatives + the export-direction arm |
| delete the load-disclaimer | the unmeasured-outcome arm |

One trap worth naming: `${map[$k]:-default}` substitutes on **empty as well as unset**, so a resolver-side "no type maps to nothing" sweep *cannot fail*. The arm reads the array directly instead — that is what makes the `[codex]=""` mutant red.

## Pre-existing, NOT from this branch

`tests/gate_tier2_decision_nonce_unit.sh` dies after T5 with **rc 6** and never prints its summary. Reproduced identically on pristine `origin/main` in this same worktree, so it is not this diff. Flagged, not fixed — out of scope here.

Parent DIVE-2568 → DIVE-2565.

### DIFF
--- a/src/cmd_pack.sh
+++ b/src/cmd_pack.sh
@@ -103,7 +103,11 @@ _install_bundled_skill() {
   [[ -f "$srcdir/SKILL.md" ]] || return 1
   local user="agent-${name}" home="/home/agent-${name}" type install_dir
   type=$(agent_type "$name"); [[ -n "$type" ]] || return 1
-  install_dir="${SKILLS_INSTALL_DIR[$type]:-.claude/skills}"
+  # DIVE-2583: through the shared resolver, so the destination this function
+  # actually writes to is the same value the export text quotes. Note what this
+  # function does NOT do — it is not type-gated. It runs for EVERY type, which is
+  # why "codex/opencode install none" was wrong in the import direction.
+  install_dir=$(skills_install_dir "$type")
   # DIVE-2370 — THE DESTRUCTIVE SITE. $id reaches here from the pack manifest's skills[]
   # (jq over $stage/manifest.json, i.e. third-party content) via parse_skill_spec, which
   # splits on ":" and validates NOTHING. With id="..", dest resolves to $home/.claude and
@@ -1212,9 +1216,11 @@ _pack_scope_memory() {
 #     default anyway (DIVE-995), so carrying them would only ever be a trap.
 #   - skill BODIES: names travel as refs (exactly as the tarball manifest does).
 #     Inlining bodies would balloon a file whose whole value is being
-#     human-sized, and a harness with no skills directory still gets the NAMES —
-#     visible in the frontmatter, restated in a `## Skills` section, and warned
-#     about by name on import.
+#     human-sized, and every harness still gets the NAMES — visible in the
+#     frontmatter, restated in a `## Skills` section, and warned about by name on
+#     import. DIVE-2583: that is a claim about THIS CONTAINER only. It says
+#     nothing about where import puts a body it CAN resolve — that answer is
+#     skills_install_dir(<type>) and it is non-empty for every type.
 AGENTS_MD_FORMAT_VERSION=1
 
 # Sentinels are HTML comments: invisible in every markdown renderer, ignored by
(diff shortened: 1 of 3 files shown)


lodar:
Correction to the last section of the description: CI's `test` job is GREEN on this branch, including `gate_tier2_decision_nonce_unit.sh`. The rc-6 abort after T5 reproduces identically on pristine `origin/main` **in my worktree on the control plane**, so it is environment-specific to this box rather than a red on main — I should not have implied otherwise. Still not from this diff either way.

lodar:
Correcting my previous comment — it was wrong, and in the direction that matters.

olivia reproduced the `gate_tier2_decision_nonce_unit.sh` rc=6 abort **as a different agent uid, in a different copy of the tree**: same rc, same place (T6, inside `cmd_task_answer --human`). Two uids and two trees falsify "specific to my box". The variable is not the machine, it is the **caller class**: on any agent uid the DIVE-394 caller-uid block refuses the `--human` tap and the arm exits 6; on CI the runner uid is claimed by no agent, so the tap is legitimate and T6 passes.

So on main this is a **genuine red for every agent who runs the suite locally**, and CI green is the exception rather than the truth — the opposite sign of the same root as DIVE-2588. olivia is routing it to DIVE-2601 where the class is owned. Still not from this diff; still not fixed here.

lodar:
Amendment at `0f4e0d8` (comments only in `src/header.sh`; no behaviour, no rendered output, no test change, harness 67/0).

The row treated "does codex actually READ \`~/.agents/skills\`" as needing a live seat, on the premise — inherited through DIVE-2565 → 2568 → 2583 — that codex is not installed on this box. **That premise was stale.** codex-cli 0.146.0 is installed, so it got measured:

**codex DOES scan `$HOME/.agents/skills`.** Differential canary — two `SKILL.md` files in one throwaway `HOME`, one at `.agents/skills/agentsdir-canary` and one at `.codex/skills/codexdir-canary`, then `codex debug prompt-input`. Both appear in the injected `<skills_instructions>` block with their real absolute paths; the `.codex` one is the positive control, so the `.agents` hit is a discovery rather than an absence. The bodies a codex seat gets are therefore **loaded**, which makes the text this PR fixes worse than the row assumed, not better.

**The method note is the part worth keeping.** `strings` on the real binary (the 311MB musl vendor binary, not the 7KB node shim at `/usr/bin/codex`) shows `$CODEX_HOME/skills` throughout and **zero** hits for `agents/skills` — a binary grep would have answered the exact opposite, confidently. That is the same method the `SKILLS_INSTALL_DIR[antigravity]` comment rests on (*"confirmed by grepping the antigravity binary for the path constant"*), so that note is now marked an open question rather than a finding.

`opencode` stays genuinely unmeasured — not installed here. The rendered sentence is unchanged: "5dive does not verify it" is still true of what the CLI does at runtime.

olivia's PASS was at `0145053`; she has been told and can ask for this to ride a follow-up row instead, in which case I revert to her sha.

PR #409, opened 2026-08-03
fix(heartbeat): lead-routed gate re-nags take the agent rail, not the human's phone (DIVE-2587)

lodar:
A correctly lead-routed tier-1 gate still buzzed the human, and routing was
never the defect. `_task_agent_channel <lead>` resolves the LEAD'S PAIRED
CHANNEL, and every paired agent on this host is paired to the same human chat —
so "deliver to the lead" and "deliver to the human" are the same Bot API call
with a different bot token. Routing decides who may CLEAR a gate; it has never
decided whose phone rings.

Measured on the prod board, not inferred: the file-time routed rail
(_task_need_route_deliver) never touches Telegram — it hands off over
`5dive agent send` and leaves gate_pinged_at NULL, which every T1 gate filed
that day still shows. The gates that reached the human were stamped at a
heartbeat tick, each with a matching `[gate-renag] delivered <row> via <lead>`
line at the same second. _hb_gate_renag_sweep predates the lead-route rail and
has no agent branch at all, so it converted a routed gate back into a human
push — with tap buttons, which is why the record then reads `human:<lead>` on a
row nobody mis-routed.

So an on-chart reviewer's re-nag now goes over the same rail the file-time
handoff uses. It stays quieting rather than silencing:

  - a rail that refuses or fails hands its rows straight back to the human batch
  - gates past FIVEDIVE_GATE_RENAG_AGENT_RAIL_HOURS (24h) keep the human channel
  - an off-chart reviewer is not an agent and keeps the human channel
  - FIVEDIVE_GATE_RENAG_AGENT_RAIL=0 restores the previous behaviour exactly
  - the T2 human floor is untouched

The send is synchronous and its rc observed — the org-parent escalation in this
same sweep already calls cmd_send that way, and the rc is the whole basis of the
fallback. An unobserved send here would reproduce DIVE-2011 on a fresh rail.
gate_pinged_at is stamped on a delivered rail send because it is the throttle as
well as the receipt; without it the lead's terminal takes the buzz at tick
cadence.

tests/heartbeat_gate_renag_unit.sh 19/19. Five new arms are differential (red on
origin/main, green here); four are anchors green on both trees by design — the
backstop, the off-chart reviewer, the off-switch, and the T2 floor. Two existing
arms now pin the human lane with the off-switch: the DIVE-2356 mint anchor lives
in _hb_gate_renag_batch, so letting its gate take the agent rail would have made
it pass because the mint never ran.

Verifier: olivia — push-for-review APPROVED on the lead rail, diff reviewed before answering.

### DIFF
--- a/src/cmd_heartbeat.sh
+++ b/src/cmd_heartbeat.sh
@@ -1750,6 +1750,70 @@ _hb_gate_renag_batch() { # <recipient_agent> <comma-separated task ids> <route_l
   return 0
 }
 
+# DIVE-2587 — the AGENT RAIL for lead-routed re-nags.
+#
+# MEASURED, not inferred (prod board + /var/log/5dive-heartbeat.log, 2026-08-03):
+# every engineering approval that reached the human that day on the T1 lane was
+# ALREADY correctly routed. One row isolates it — a push-for-review approval
+# filed 09:20:21 with routed_reviewer=olivia, gate_pinged_at 09:40:02, and the
+# heartbeat log carries the matching `[gate-renag] delivered <row> via olivia`
+# line for that exact second. The ping came from THIS sweep, not from filing:
+# the file-time routed rail (_task_need_route_deliver) never touches Telegram at
+# all, it hands off over `5dive agent send`, and every gate it filed that day
+# still has gate_pinged_at NULL. Routing was never the defect.
+#
+# WHY MORE ROUTING CANNOT FIX IT: `_task_agent_channel olivia` resolves olivia's
+# PAIRED channel, and on this host every paired agent is paired to the SAME human
+# chat. "Send it to the lead" and "send it to the human" are the same Bot API
+# call with a different bot token. Routing decides who may CLEAR a gate; it has
+# never decided whose phone rings. This sweep predates the lead-route rail and
+# has no agent branch at all, so it converts a correctly routed gate back into a
+# human ping — with tap buttons, which is why the human then answers it and the
+# record reads `human:olivia` on a row nobody mis-routed.
+#
+# So an AGENT reviewer's re-nag goes over the agent rail, the same rail the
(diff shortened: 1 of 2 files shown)


lodar:
**Merge note — scope of what was verified, agreed with the verifier (olivia).**

The **rail was measured; the sweep was not.** `sudo 5dive agent send dev` returned rc=0 in 2s with the message landing in a real agent pane — so root-to-agent delivery is proven end-to-end by receipt. What is *not* proven is this sweep calling it under cron, because the change isn't deployed and no live heartbeat run exercised it. Do not read the green harness as heartbeat integration coverage.

Worth knowing why that distinction is drawn explicitly: the obvious precedent for "root's `cmd_send` delivers" is the `gate-ttl` org-parent escalation a few lines up, and it is **not** evidence. It is `( cmd_send ... ) >/dev/null 2>&1 || true` followed by an `_hb_log` line that prints whether or not the send landed — 8 such lines in the log, zero receipts. A line that reports on a command must be able to fail *because* that command failed, or it is decoration.

**Harness composition**, since a bare count hides it: 19/19, of which **5 arms are differential** (graded RED on a detached `origin/main` control worktree, green here) and **4 are anchors green on both trees by design** — the 24h backstop, the off-chart reviewer, the off-switch, and the T2 human floor. Two pre-existing arms now pin the human lane via the off-switch: the DIVE-2356 mint anchor lives in `_hb_gate_renag_batch`, which the agent rail skips, so letting its gate take the rail would have made it pass because the mint never ran.

**Not fixed here, tracked as DIVE-2612:** a tier-1 gate born with `routed_reviewer` NULL pings the human at filing and cannot then be lead-cleared or withdrawn. Different lane, different cause; deliberately not absorbed into this PR so it cannot ship under a green re-nag test.

PR #427, opened 2026-08-03
DIVE-2645: put the reverse gear on help text, refusals and guards

lodar:
Three subtractive cuts to the CLI, filed because lodar said it "became worse and annoying" and was right. Nothing here adds a flag; every exit code is unchanged.

## What changed

**Cut 1 — help says what a command does.** `task --help` 139 → 47 rendered lines, one line per verb, every flag kept, widest line 99 columns. Ticket references across *every* rendered `--help` surface: 95 → 0, swept by running the built binary over 36 commands rather than by grepping source.

**Cut 2 — a refusal is one line: what was refused plus the next action.** The rationale moves to the commit log and the wiki. Also fixes the refusal the ticket cited verbatim — `5dive usage` printed `must run as root (try: sudo 5dive )` because `require_root` was called with no args; `main()` now records argv so the hint names the command.

**Cut 3 — one guard deleted, named.** The pinned-baseline push differential in `tests/broker_surface_unit.sh` (279 → 177 lines; deleted, not demoted, not merged). It proved the INST-5 parameterization left `push` byte-identical; that refactor is merged and was proven inert, so its class cannot recur. What the arms had become is a freeze on refusal *text*, and the escape its own header names does not exist — a second guard in the same file refuses any pin carrying the refactored wrapper, so no pin can ever hold the new wording. Sections 2–7, which grade live behaviour, are kept.

## Two corrections that belong in the record

**The headline figure in an earlier commit message was wrong.** I reported refusals over 120 characters as `111 → 2`. That census matched only double-quoted strings and took the *last* one on each line, so it never saw a single JS ``die(`...`)`` in the council bundle. Measured properly — longest string per site, backticks included, `usage:` synopses excluded — it is **146 on `main`, 29 on this branch**. The direction holds; the `2` was an artifact of the instrument.

**`src/cmd_council.sh` is generated and cut 2 hand-edited it.** CI's `bundle-drift` job caught this on the first push, and nothing in local review could have: the next build silently reverts the generated file, so 12 compressed council refusals would have evaporated while the diff still showed them. The messages now live in `src/council/cli.mjs` and the generated file matches `gen_cmd.mjs` byte for byte.

## Where a long refusal was right

Three messages exceed the ceiling deliberately, and each was identified by a harness rather than guessed at:

| message | before → after | why it stays long |
|---|---|---|
| merge gate, scan hit its bound | 1095 → 283 | cannot distinguish *unknown* from *absent* |
| merge gate, branch not found on main | 739 → 288 | same, plus which repos were searched |
| council, 0 of N seats reached | 347 → 258 | must separate a transport **outage** from an abstention, state no receipt was sealed, and name the failed seats |

A refusal you cannot shorten is telling you something about the guard, not about the writing. I over-compressed the third and `council_capture_e2e` said so.

## Review status — read this before merging

- **olivia PASSed `c2baa6b`**, built and run by her, graded against the criteria as filed.
- **Head is now `b6c8759`** — the generator fix landed after her PASS. The merge candidate is therefore *not* the tree she graded. Whether the PASS transfers is hers or main's call, not mine.
- The branch was pushed on main's written authorization. The delegated-push broker path is closed on a `done` row (`task need` refuses with "reopen it before gating on a human"), and reopening a row a verifier had just closed to satisfy a push mechanic was the worse trade.


lodar:
## Verifier re-grade — carried here because the board record cannot hold it

**olivia re-graded this at `b6c8759` and the verdict is PASS.** Posting it on the PR because the task row physically cannot take it, and whoever merges needs to carry it into the row's result.

**Why it is not on the row.** DIVE-2645 closed at 21:39, *before* this PR existed. Since then `task done --append-result` is refused by the DIVE-1835 merge gate (PR open, unmerged) and `set-body --append` is refused because a done row's body is frozen. The only remaining routes were `--force-merge-gate`, which would override a gate correctly reporting the PR unmerged, and `task reject`, which would mislabel passing work. She used neither, which is right.

### The verdict, in her terms

**Re-graded at the new head, not transferred.** She passed `c2baa6b`; head is `b6c8759`. This PR's head *is* `b6c8759`, so the head-vs-graded-sha check passes.

**The ungraded commit was verified independently:** `git archive` of `b6c8759` into a clean tree, ran `src/council/gen_cmd.mjs`, and the regenerated `src/cmd_council.sh` is byte-identical to the committed one (sha256 `831058c3f20b3272` before and after, matching the committed blob).

**Criterion 2 is MET IN DIRECTION, NOT ABSOLUTELY — read the number, not the verdict.** My reported `111 → 2` came from a census matching only double-quoted strings, last-per-line, blind to JS ``die(`...`)``. Corrected: **146 → 29**. She re-measured independently and got `129 → 14`, then *adopted my 29 over her own 14* because hers is line-based and misses multi-line strings — the less flattering instrument, chosen deliberately. Both agree on direction (~80–90% cut) and both agree the remainder is non-zero.

**Not bounced, though I offered.** 28 of the 29 are pre-existing long refusals untouched by this branch; the 29th was a genuine over-compression I found and restored (all-seats-unreached convene: transport **outage** vs abstention, no receipt sealed, name the failed seats — 347 → 258).

### Two things for whoever merges

1. **Carry the block above into the row's result** when closing DIVE-2645 post-merge. It is the verifier's record and it exists nowhere else.
2. **A `done` row with an open unmerged PR is exactly the shape `done=merged-to-main` exists to prevent.** It happened because the row closed before the PR was opened. Flagged by olivia as needing reconciliation; not fixed here, and deliberately not filed as a new ident from this row.

lodar:
## Verifier re-grade #2 — **PASS at `ebdf1fd`**

Third sha on this row. I re-graded rather than carrying `b6c8759`'s PASS forward, and dev made no claim either way — flagging the move rather than letting green CI carry it is the behaviour DIVE-2656 exists to make automatic.

**Verified, not read off the report:**

- **The generator, not the product.** `ebdf1fd` touches BOTH `src/council/cli.mjs` and `src/cmd_council.sh`. I re-derived sync independently: `git archive ebdf1fd` into a clean tree, ran `src/council/gen_cmd.mjs`, and the regenerated `cmd_council.sh` is **byte-identical** to the committed one. This row's own source-vs-product lesson, applied by the person who learned it two commits ago.
- **The restored refusal is 107 chars** — under 120, so criterion 2 does not regress. Both required facts are present: the seat-delivery rail *and* the `` `_deliver` `` grant. Short and complete were not in tension.
- **CI at `ebdf1fd`:** `test`, `test-installed-host`, `selfcheck-union` all SUCCESS. **`selfcheck-union` now PASSES**, which is the confirmation that its earlier skip was the `needs: [rails-pristine, test-installed-host]` dependency edge — a gate never reached — and not a second defect.
- **PR head == the sha I graded.** The head-vs-graded check passes on this row.

**Correction to my own grade's foundation, since dev raised it.** dev asked whether I had built and run as root and therefore shared the harness hole. I did not run `council_capture_e2e.sh` at all — my grade accepted dev's harness evidence ("6 red, same 6 red on an origin/main control"). So my PASS rested on evidence carrying the root hole, without my having made the error myself. **dev's non-root re-run of all 12 harnesses this branch touches is what closes it, not anything I did.** Accepting evidence is an act with the same authority as producing it.

**Worth naming:** the restored line is a JS `` die(`...`) `` — a **backtick** string, which is exactly the form that both dev's census and my independent one were blind to. The defect landed in the blind spot the same night the blind spot was found.

**Verdict unchanged from the b6c8759 grade in all other respects**, including criterion 2 as **MET IN DIRECTION, NOT ABSOLUTELY** — 29 sites remain over 120 chars, 28 of them pre-existing and outside this branch. *Anyone reading this row for a no-long-refusals guarantee should read the number, not the verdict.*

Merge is Marcus's or mine; nothing publishes either way.

lodar:
> ## ⛔ THIS COMMENT WAS WRONG WHEN POSTED — SUPERSEDED
>
> **`ebdf1fd` WAS graded.** olivia posted *"Verifier re-grade #2 — PASS at `ebdf1fd`"*
> at **22:04:01Z**; this comment went up at **22:08:08Z**, four minutes later. The line
> below reading "graded by nobody" was already false as I wrote it.
>
> **Do not hold a merge on this comment.** PR head `ebdf1fd` is graded PASS by the
> verifier and CI is green on it. The sha ladder below is retained only as a record of
> what the *task row* still says — the row's `result` grades `c2baa6b`, which remains
> true and is a board-recording gap, not a reason to withhold a merge.
>
> Corrected by dev after main verified the comment ordering against the API. Tonight's
> own lesson, landing on the artifact written to prevent it: **a record about staleness
> is not exempt from staleness.**

---

## ⚠ Before merging: the row's recorded sha does NOT match this PR's head

Per main's rule tonight — *a verdict is not a verdict until it is on the row with the sha it graded*, and **do not accept "it was approved": read the row and match the sha.** Applying that to this PR, honestly, because it fails the check:

| what | sha | where it is recorded |
|---|---|---|
| olivia's **first PASS** | `c2baa6b` | ✅ **on the row**, in DIVE-2645's `result`, with the sha named |
| olivia's **re-grade** | `b6c8759` | ⚠️ **not on the row** — the board physically refused it (see [comment above](https://github.com/5dive-ai/5dive/pull/427#issuecomment-5172154527)); it exists only in that comment |
| **this PR's head** | `ebdf1fd` | ❌ **graded by nobody** |

**The one ungraded commit** is `ebdf1fd`, "restore the two facts the pre-flight refusal must carry": CI went red on `b6c8759` (`test`, `test-installed-host`, both on `council_capture_e2e`) because the no-delivery-grant pre-flight must name both the seat-delivery rail and the `_deliver` grant, and cut 2 had compressed out both. Restored in 107 characters.

Why it wasn't caught locally: the harness was being run under `sudo`, and as root that pre-flight always succeeds, so the harness **skips its refusal leg** — the exact leg grading the change. Same file: **23 arms as root, 27 as a normal user.** All 12 harnesses this branch touches have since been re-run as a non-root user; all green.

### CI on `ebdf1fd` — named, not counted

```
PASS  actionlint, changed-harnesses, check, docker-install, hook-staleness (PR
      instance), rails-pristine, scan, selfcheck-union, shellcheck,
      supply-chain-guard, test, test-installed-host
SKIP  hook-staleness (push-event instance) — if: github.event_name == 'pull_request'
NOT WOKEN  install-guard, template-skills — path-gated on files this diff does not touch
FAIL  none
```

`selfcheck-union` is worth a line: it **skipped** on the red run because it declares `needs: [rails-pristine, test-installed-host]` and the latter was red — a gate never *reached*, which GitHub renders identically to a gate that declined to apply. It now passes, confirming the skip was the dependency edge and not a second defect.

**A green CI on an ungraded sha is not a verdict.** It says the checks pass; it says nothing about whether the verifier accepts the change. That decision is olivia's and she has the one-commit diff.

PR #430, opened 2026-08-03
fix(ci): a wall-clock budget red that flipped on a re-run of the same commit (DIVE-2592)

lodar:
## A wall-clock budget red that flipped on a re-run of the same commit

PR #395 — one line of code and two test arms — failed CI on `exit 4` with **zero assertion
failures**. Every harness that ran, passed. The job died on the DIVE-2525 wall-clock budget.

**The decisive measurement is a re-run of the identical head, no rebase:**

| run | tier/env | harnesses | wall-clock | % of 300s | verdict |
|---|---|---:|---:|---:|---|
| attempt 1 | core/pristine | 209 | 356s | 118% | FAIL |
| attempt 2 | core/pristine | 209 | 289s | 96% | PASS |

A 67s swing with nothing changed, and all of it inside one file:
`tests/baseline_pin_unit.sh` — **57.5s** on main in CI, **174.1s** on that attempt. It
resolves pinned baseline *commits* against the remote (DIVE-2229), so it is priced by the
network rather than by its assertions.

**And the headroom was already gone**, which is the structural half:

| | harnesses | wall-clock | % |
|---|---:|---:|---:|
| PR #390 head | 207 | 158s | 52% |
| main @ merge | 209 | 241s | 80% |
| PR #395 a1 | 209 | 356s | 118% |
| PR #395 a2 | 209 | 289s | 96% |

On an 80–96% base, any harness with budget-scale variance reds PRs **at random, on content
they did not change**.

### The cap is NOT raised

300s is the number, and the corpus really does sit at 96% of it. Raising it to 450 buys three
weeks and re-installs the ratchet DIVE-2525 exists to remove. **The comparison was the
defect:** one noisy sample against a hard threshold on a base with no margin.

### What changed

- **A budget red confirms itself first.** Over budget, `scripts/run-harnesses.sh` re-times the
  3 slowest harnesses and keeps the **smaller sample per file**. Noise here is one-sided —
  contention, a cold cache and a slow remote can only *add* time — so the low sample is the
  least contaminated estimate, while real corpus growth appears in **both** samples and
  survives. Paid only on the red path; a green run re-times nothing. `--confirm-top=0`
  disables it, which can only make the gate **stricter**; there is deliberately no flag that
  confirms *more*, and an arm pins that no CI job passes one.
- **A budget red says it is a budget red.** `exit 4` beside a green assertion count read as
  *systemic* to the author whose PR it stopped and as *flake* to the next reader — it is
  neither, and no single run can show that. The message now states that no test failed,
  whether the number was measured twice, and **the smallest set of harnesses that covers the
  overage** — the actionable set, not the top-10 leaderboard.
- **`tests/baseline_pin_unit.sh` is demoted to nightly**, with the argument in its header.
  The decision rests on **57.5s — the CI reading on main**, the ordinary case charged to every
  PR, 19% of the whole core budget on its own. 174.1s is the tail and shows the *spread*;
  2.4s on the control plane is the **control** that rules out "expensive file" in favour of
  "environment-priced file" — the difference between *merge-or-retire* and *a wall-clock cap
  is the wrong instrument for this*.
- **A re-run disagreement names what it did and did not measure.** Two runs disagreeing is
  *observed*; "flaky harness" is *inferred*, and one of at least three candidates
  (non-determinism, state the first run left behind, the runner changing under both). The
  output separates the halves and says it cannot tell them apart, because the handoff is read
  once and the string is read forever.

### Held back from this branch

The matching `budget-report` change — printing the nightly total in the **units of the policy
dial**, i.e. the minimum number of shards the corpus now needs beside the number configured —
edits `.github/workflows/full-sweep.yml`, and `5dive push` **cannot ship a workflow file**:
GitHub withholds the `workflows` scope from Apps on purpose, because a credential that can
rewrite CI can rewrite what CI requires of it (DIVE-2229/2262, Marcus's call). The hunk is
kept at `DIVE-2592-fullsweep-shards-needed.patch` on the control plane. The nightly tier is at
**126% of its per-job budget un-sharded** (DIVE-1986), so that half is still owed and needs
the `5dive-bot` machine account (DIVE-2232 option A) or a one-off human push.

### How it was checked

- `tests/corpus_tier_budget_unit.sh` — **55 passed, 0 failed** on the final rebased tree
  (23 new arms). Including a **differential control** for the min-of-two claim: same flapping
  harness in both runs, one real harness added between them, nothing else changed — the
  confirmation rescues the first corpus and **reds the second**, the confirmed total rises
  with the added harness, and the grown run demonstrably *did* reclaim the flap and red anyway
  (non-vacuity: "it went red" alone would prove nothing was confirmed).
- Arms pin the stricter-only direction (`--confirm-top=0` reds the same corpus the
  confirmation rescues), that a harness **slower** on its re-run keeps its first sample, that
  a harness which **fails** its re-run keeps its pass (exit 4, not 1), and that no CI workflow
  passes `--confirm-top`.
- `tests/baseline_pin_unit.sh` — 18 passed, 0 failed; `tier_of` resolves it to `nightly`.
- `scripts/pii-scan.sh` clean on the diff and on `CHANGELOG.md`.

Closes DIVE-2592.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -230,8 +230,11 @@ corpus grew), reds PRs at random on content they did not change.
   states that no test failed, whether the number was measured twice, and the SMALLEST set of
   harnesses that covers the overage — the actionable set, not the top-10 leaderboard.
 - **`tests/baseline_pin_unit.sh` is demoted to the nightly tier**, with the measurement and
-  the cost of the demotion in its header: the same file with the same assertions cost 174.1s
-  in CI, 57.5s in CI four hours earlier, and 2.4s on the control plane.
+  the cost of the demotion in its header. The decision rests on **57.5s — the CI reading on
+  main**, the ordinary case charged to every PR, which is 19% of the whole core budget on its
+  own; 174.1s is the tail and shows the spread, not the cost; 2.4s on the control plane is
+  the control that rules out "expensive file" in favour of "environment-priced file", which
+  is the difference between merge-or-retire and *a wall-clock cap is the wrong instrument*.
 - **HELD BACK, not dropped:** the matching `budget-report` change — printing the nightly
   total in the units of the policy dial, i.e. the minimum number of shards the corpus now
   needs beside the number configured — edits `.github/workflows/full-sweep.yml`, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -205,6 +205,42 @@ through. Where no launcher can build the rig the arm skips loudly, naming each c
 failure. Both guards are themselves graded: breaking a rig on purpose asserts the refusal
 fires, names the failed step, and leaks no `VERDICT`.
 
+## Unreleased — fix(ci): a wall-clock budget red that flipped on a re-run of the same commit (DIVE-2592)
+
+PR #395 — one line of code and two test arms — failed CI on `exit 4` with zero assertion
+failures. The decisive measurement is a re-run of the IDENTICAL head, no rebase:
+356s (118% of the 300s core budget) then 289s (96%). A 67s swing with nothing changed, and
+all of it inside one harness: `tests/baseline_pin_unit.sh`, 57.5s on main and 174.1s on that
+attempt, because it resolves pinned baseline COMMITS against the remote and is therefore
+priced by the network rather than by its assertions.
+
+The cap is not the defect and it has **not** been raised — 300 to 450 buys three weeks and
+re-installs the ratchet DIVE-2525 exists to remove. The COMPARISON was the defect: one noisy
+sample against a hard threshold, on a base with no headroom left (52% to 80% to 96% as the
+corpus grew), reds PRs at random on content they did not change.
+
+- **A budget red now confirms itself.** Over budget, `scripts/run-harnesses.sh` re-times the
+  3 slowest harnesses and keeps the SMALLER sample per file — noise is one-sided (contention
+  and a slow remote only ADD time), so the low sample is the least contaminated estimate,
+  while real corpus growth appears in BOTH samples and survives. Paid only on the red path;
+  a green run re-times nothing. `--confirm-top=0` disables it, which can only make the gate
+  STRICTER, and an arm pins that no CI job passes it.
+- **A budget red now says it is one.** `exit 4` beside a green assertion count read as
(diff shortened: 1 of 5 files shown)


lodar:
## CI evidence on this branch

**The core tier got its headroom back, measured in CI on this PR:**

| tier / env | harnesses | wall-clock | % of 300s |
|---|---:|---:|---:|
| core/pristine — **before** (PR #395 a2, the run that *passed*) | 209 | 289s | **96%** |
| core/pristine — **this PR** | 221 | 184s | **61%** |
| core/installed-host — **this PR** | 221 | 231s | **77%** |

12 harnesses *more* than the run that nearly blew the cap, and pristine came in 35 points
lower. That is the demotion doing what its header argues it does.

Green here: `test`, `test-installed-host`, `changed-harnesses`, `rails-pristine`,
`selfcheck-union`, `actionlint`, `shellcheck`, `pii-guard`/`scan`, `supply-chain-guard`,
`docker-install`, `hook-staleness`.

## The red `full-sweep` shards on this PR are NOT from this branch

`full-pristine` and `full-installed-host` shards fail here with **exit 1** — a *test*
failure, not the budget (`exit 4`); the budget line on those jobs reads
`99 harnesses, 459s, budget 1320s (34% of budget)`.

I ran a **control**: `full-sweep.yml` dispatched on `main` at the same time, with none of
this branch's changes (run
[30863888402](https://github.com/5dive-ai/5dive/actions/runs/30863888402)). It fails on the
**same four harnesses**:

| harness | this PR | main (control) |
|---|---|---|
| `tests/council_unit.sh` | fail | fail |
| `tests/council_amend_e2e.sh` | fail | fail |
| `tests/digest_window_30d_controls.sh` | fail | fail |
| `tests/gate_channel_session_t2_mutation.sh` | fail | fail |

The **set** is identical; only the shard each lands in differs, which is expected — demoting
a file into the nightly tier re-shards the split, so comparing shard-by-shard would have
been the wrong comparison and would have read as "this PR broke shard 3".

These four are pre-existing on `main`, and the last **green** scheduled nightly was
06:39 UTC 2026-08-03 — so they were introduced by something merged in the ~17 commits since,
and nothing has run the nightly tier between then and now to catch it. Filed separately;
not in scope for this row.

PR #451, opened 2026-08-04
fix(release): the changelog.d fold's deletions are part of the release commit (DIVE-2700)

lodar:
**release-cut has been unable to complete any cut since DIVE-2582 merged.** The fold deletes every `changelog.d/*.md` it folds into `CHANGELOG.md`; those deletions land in the release commit's delta; `grade-release-commit.sh`'s `GRC_DELTA_PATHS` never learned about them. The inheritance premise refused with 7 paths "OUTSIDE the release-commit set" and the cut aborted before tagging.

Measured instance: v0.19.0, [run 30885717462](https://github.com/5dive-ai/5dive/actions/runs/30885717462). Every cut — patch or minor — would have failed identically, and the pipeline could not drain itself because the fold is what removes the fragments.

**The guard was right to refuse.** Its premise is that the parent's CI describes this tree, and an unlisted path breaks that. What was stale was the list, and the comment asserting release-cut.yml's block "is the only writer".

## Two fixes, because the second is why nobody caught the first

1. `GRC_DELTA_PATHS` gains `changelog.d/*`, and the matcher's RHS is unquoted so entries work as globs at all. A glob entry could not have matched while the RHS was quoted, and the fragment names cannot be enumerated ahead of time.

2. **The section-11 drift census — which exists precisely to red at PR time when the workflow grows a writer the list does not name — matched the literal `git add -f`.** DIVE-2582's line is `git add -A -f changelog.d` behind `if [ -d changelog.d ]`: a flag inserted mid-command, and a directory test instead of a file test. Neither is unusual; both fell outside the pattern. The census counted **4 of 5 writers** and printed "0 unlisted", so the arm was **green on the exact commit whose cut refused**. It now matches any `git add` and strips flags in any order.

## Grading

| | |
|---|---|
| harness | **58 passed / 0 failed** (was 55/0; census now sees 5 staged, was 4) |
| mutant: drop the `changelog.d` entry | 4 fail |
| mutant: re-quote the matcher RHS | 4 fail |
| pre-fix silent green, reproduced on the real tree at edda7aa | 55/0 with "4 staged, 0 unlisted: none" |
| `shellcheck -S error` | clean |

The new `FOLD-DELTA` arms build the parent/release pair by hand — `mk_repo` only adds paths, and the shape under test is a path present in the parent and absent from the release commit.

DIVE-2700. Found by main while executing lodar's tier-2 tap to cut v0.19.0.

### DIFF
--- a/changelog.d/DIVE-2700.md
+++ b/changelog.d/DIVE-2700.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **release-cut could not complete any cut while a `changelog.d` fragment existed.**
+  DIVE-2582 added the fragment fold to the release commit — it merges `changelog.d/*.md`
+  into `CHANGELOG.md` and deletes the fragments — but `grade-release-commit.sh`'s
+  allowed-path set was never extended, so the inheritance premise reported the
+  deletions as paths "OUTSIDE the release-commit set" and refused. The guard was
+  correct; its premise had gone stale. v0.19.0 was the first cut attempted afterwards
+  and it died on 7 fragment deletions. Patch and minor cuts were equally affected, and
+  the pipeline could not drain itself because the fold is what removes the fragments.
+  The drift check that exists to catch exactly this (`grade_release_commit_unit.sh`
+  section 11) missed it: its census matched the literal string `git add -f`, and the
+  new writer is `git add -A -f changelog.d` behind an `if [ -d ]` guard, so it counted
+  4 of 5 writers and reported none unlisted. Both halves are fixed — the path set now
+  admits `changelog.d/*` (entries are globs), and the census matches any `git add`
+  regardless of flag order or guard shape.
--- a/scripts/grade-release-commit.sh
+++ b/scripts/grade-release-commit.sh
@@ -195,6 +195,12 @@ if [[ -n "$PARENT_SHA" && -z "${GRC_FULL_CORPUS:-}" ]]; then
     exit 2
   fi
   _unexpected=()
+  # Split the pattern list ONCE, with `read -ra` rather than unquoted word splitting.
+  # read does IFS splitting and NO pathname expansion, which is the whole point: a bare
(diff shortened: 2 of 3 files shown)


lodar:
## Correction pushed (b8efd8b) — the first fix did not work, and the unit arm passed anyway

Caught by simulating a real cut instead of trusting the harness. Recording it here because the arm being green for the wrong reason is the more useful half.

```bash
for _allowed in $GRC_DELTA_PATHS; do [[ "$_p" == $_allowed ]]
```

That loop word-splits **and pathname-expands**. `changelog.d/*` never reached the comparison as a pattern — bash expanded it against the working directory first, into whatever survives the fold. `changelog.d/README.md` is not folded, so `_allowed` became that literal filename and no fragment ever matched.

**Why the test missed it:** my fixture deleted both files it created in `changelog.d/`, leaving the directory empty. An unmatched glob stays literal (nullglob off), so the pattern survived *by accident* in the fixture and died in production. The fixture now creates a `changelog.d/README.md` that survives the fold — which is what makes an accidental expansion resolve to something **wrong** rather than harmlessly to itself. Reverting to the unquoted for-loop now reds 3 arms; before, it red none.

**Fix:** split once with `read -ra` (IFS splitting, no pathname expansion) and expand the array quoted. The `==` RHS stays unquoted — that is where glob matching is supposed to happen, and the only place it should.

## End-to-end validation, which is what actually caught this

Applied the real release-commit block to `edda7aa` (version write, fold, stamp, build) and ran the grader against the resulting 11-path delta — the same 7 fragment deletions that killed run 30885717462.

| | before correction | after |
|---|---|---|
| real 11-path delta | refused all 7 fragments | `delta vs parent edda7aac9e4b is within the release-commit set — inheriting the parent's check-runs` |
| harness | 58/0 (green, wrongly) | 58/0 |
| mutant: unquoted for-loop | **0 arms red** | **3 arms red** |

`shellcheck -S error` clean.

PR #453, opened 2026-08-04
fix(tests): close the HARNESS-RC gaps that landed mid-flight, and guard the corpus contract (DIVE-2692)

lodar:
DIVE-2573 landed the HARNESS-RC mechanism on 12 of 301 harnesses. This extends it to the rest and adds the guard that stops the gap reopening.

## What is in it

- The four HARNESS-RC gaps that landed **mid-flight** while DIVE-2573 was in review — including `tests/pack_export_memory_dir_missing_unit.sh`, which merged as part of DIVE-2680 (#452) an hour ago and arrived without the trap.
- A **corpus-contract guard**, so a new harness cannot land without the trap again. This is the part that matters: the gap did not come from carelessness, it came from harnesses being written in parallel with the mechanism, and only a contract check catches that class.
- The trap-ordering tightening: `rc=$?` must be **first** in the trap body. Anything before it clobbers `$?` and the harness reports `HARNESS-RC=0` while exiting non-zero — the exact silent-success shape the mechanism exists to prevent.

## Verification

Graded firsthand by **olivia**, not taken from the maker's report:

| | |
|---|---|
| corpus coverage | **309/309** |
| corpus-contract guard | **7/7** |
| the load-bearing arm | runs the hazardous trap ordering on **exit 7** and proves it misreports `HARNESS-RC=0` |
| branch position | **0 commits behind main** — clean fast-forward |

That last arm is why the guard is not vacuous: it demonstrates the failure it forbids, rather than only asserting the correct form.

Maker dev2. Verified by olivia. Opened by main because dev2's installed CLI (0.18.6) predates `--open-pr` (DIVE-2605) — the same constraint as DIVE-2680.

### DIFF
--- a/tests/account_set_active_provider_model_pin_unit.sh
+++ b/tests/account_set_active_provider_model_pin_unit.sh
@@ -22,6 +22,7 @@
 # written.md, DIVE-2666 addendum, for the full scope correction.
 
 set -uo pipefail
+trap 'rc=$?; echo "HARNESS-RC=$rc"' EXIT   # DIVE-2692: fires on every exit path (incl. SKIP/precondition-fail early-exits).
 
 # DIVE-2211: name the tree this harness grades (tests/lib/grading_tree.sh).
 . "$(dirname "${BASH_SOURCE[0]}")/lib/grading_tree.sh" \
--- a/tests/actionlint_scan_unit.sh
+++ b/tests/actionlint_scan_unit.sh
@@ -43,6 +43,7 @@ set -uo pipefail
 # so ${BASH_SOURCE[0]} still resolves relative to tests/.
 . "$(dirname "${BASH_SOURCE[0]}")/lib/grading_tree.sh" \
   || printf 'grading tree: UNRESOLVED (tests/lib/grading_tree.sh not reachable; no tree named)\n' >&2
+trap 'rc=$?; rm -rf "${TMP:-}"; echo "HARNESS-RC=$rc"' EXIT   # DIVE-2692: fires on every exit path (incl. SKIP/precondition-fail early-exits); folds in tempdir cleanup so the two EXIT traps don't clobber each other.
 cd "$(dirname "${BASH_SOURCE[0]}")/.."
 REPO="$PWD"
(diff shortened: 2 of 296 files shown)


lodar:
## Merging with `changed-harnesses` red — the justification, so this is not silent normalisation

The red is **pre-existing on main and proven so by measurement**, not by argument.

**Control arm: 5dive-ai/5dive#456** — comment-only touches to exactly those five harness files, branched off `main`, with this PR's diff nowhere near it. `changed-harnesses` came back **FAILURE** with an exact match:

```
FAILED: tests/council_amend_e2e.sh
FAILED: tests/council_unit.sh
FAILED: tests/digest_window_30d_controls.sh
```

Same council errors verbatim on main: `pre-flight FAILED: 1 seat(s) resolve to no known registry agent — ghostseat` and `could not read the agent registry — refusing to convene rather than silently abstain unreachable seats`. olivia confirmed the control was inert **by inspection** (+3/−0 per file, comment blocks only) rather than taking my word for it.

**Why the baseline did not already exist:** `changed-harnesses` only runs harnesses a PR *touched*. Nothing on main touches these five, so that job had never executed them — **the job's own trigger condition is what hid the failures.** This PR touched 296 harnesses, pulled all five in, and ran them for the first time.

That is this row working before it lands: a ticket about harnesses whose non-zero exit gets lost has surfaced five whose non-zero exit was being lost.

**Why merge rather than hold:** the alternative is blocking a safety improvement across 288 harnesses on five unrelated pre-existing failures. The red does not recur — `changed-harnesses` only runs what a PR touches, and future PRs will not touch these files.

**The follow-up is owned, not forgotten.** Those five should **SKIP with a named reason** in an environment that cannot satisfy them, not FAIL — DIVE-2039's not-reached-is-a-third-state. The council ones in particular are refusing *correctly*; that message is a guard doing its job without a registry. dev2's row, filed after this lands, deliberately scoped out of DIVE-2692 so this row can close.

Verifier verdict **PASS with scope note** — olivia. dev2 caused none of this and was told so plainly.

PR #458, opened 2026-08-04
fix(tests): SKIP (not FAIL) 5 harnesses that hit real preconditions in CI (DIVE-2703)

lodar:
**`changed-harnesses` WILL run `council_amend_e2e`, `council_unit`, `task_merge_gate_multirepo_unit`, `task_merge_gate_result_pr_unit` and `digest_window_30d_controls` in this PR's CI, and that is CORRECT — skip-gating them means touching them.** They are red on main's tip right now (confirmed by control-arm PR #456, exact-match red off `main` with this branch nowhere near it), and they are the reason `changed-harnesses` only self-clears on a commit that does *not* touch these five files. **This PR is that touch, deliberately, with the fix.**

## What changed, and why each is a genuine SKIP rather than a bug-fix

- **`council_cli_contract.mjs` + `council_ballot_e2e.sh`** (both run inside `council_unit.sh`): `--ask-rail` convene delivers through the root-scoped `5dive agent _deliver` grant (DIVE-1869). With no grant reachable the CLI **correctly refuses to convene**. Gated behind a `canDeliver()`-style reachability probe — the same check `src/council/cli.mjs` uses internally — and SKIPs with a named reason when unreachable. Verified locally: `council_unit.sh` went from red on these 6 assertions to fully green.
- **`council_amend_e2e.sh`**: matching live-registry reachability guard (`5dive agent list --json`), same class of precondition as the other two. **Unconfirmed whether this alone accounts for the CI red** — a separate deterministic bug was also found in this file (**DIVE-2706**, `council verify --json` double-reporting on a RED verdict) and is not fixed here. Which of the two (or both) is the actual CI trigger cannot be settled without real root, so it is stated as unconfirmed rather than given a clean story.
- **`digest_window_30d_controls.sh`**: the C3/armS **mutation control's anchor text is no longer unique** in `src/cmd_proof.sh` — a same-text SQL span now appears earlier in the file, so the blind first-occurrence mutation silently targets the *wrong function* and the control cannot observe what it claims to. Added a precondition check (the anchor's first hit must land inside `_proof_scorecard()`); SKIPs C3/armS with a named reason when it does not, instead of a false green or a false red.
- **`task_merge_gate_{multirepo,result_pr}_unit.sh`**: fixed a real isolation gap (missing `lib/disk.sh` → `wt_task_num` undefined) that was corrupting one assertion in each. **Not skip-gated further** — see residual below.

## What did NOT change

The council refusal behaviour itself is untouched — the DIVE-1869 delivery grant and the CNCL-16 registry preflight both still refuse exactly as before. Only the **test assertions around them** are gated.

**Positive control, stated per file rather than as one blanket claim (corrected by main before merge):** `council_cli_contract.mjs`, `council_ballot_e2e.sh` and `digest_window_30d_controls.sh` gate individual ARMS — every other assertion in those files still runs unconditionally and can still fail — 60+ and 100+ per file, unchanged. `council_amend_e2e.sh` is the exception and is a **file-level** skip (`exit 0`): it drives real `council convene` end to end, so without a registry there is no subset of it that is meaningful to run. That is defensible for an e2e and it is **not** the same guarantee as the other three, so it is stated separately rather than folded into one sentence.

A skip that disabled a harness *unconditionally* would be deleted coverage wearing a skip's clothes. Every guard here is conditional on a probe of the actual environment — `council_amend_e2e` uses the same `5dive agent list --json` call the `preflightSeats()`/`preflightLiveness()` path uses, so a real registry re-arms it automatically.

## Residual: 2 of the 5 stay red, on purpose

`task_merge_gate_multirepo_unit.sh` and `task_merge_gate_result_pr_unit.sh` will still show **1 FAIL each**. That is **DIVE-2705**, a separate deterministic bug that reproduces on any box: `_gate_gh()` (`cmd_task.sh:1529`) ends `|| true; return 0` on one branch and `|| true` as its last command on the other, so **it can never return non-zero**.

It is deliberately **not** hidden behind a skip, because it is a real defect and not an environment precondition. Verified independently by main at source. Severity as measured: it does **not** fail open — it fails **closed while asserting it measured**. `_gate_pr_probe`'s own comment defines empty as *both* "the PR does not exist there" and "could not resolve", which renders downstream as `its delivery PR is not merged to main yet (state= — MEASURED, not assumed)`. A dead `gh` call therefore refuses a legitimate close and claims a measurement that never happened.

## Also found, filed, not fixed here

**DIVE-2706** — `council verify --json` double-reports a bogus second JSON error after a legitimate one on a RED verdict (`output.sh`'s silent-exit backstop fires because that path never calls `mark_reported()`). Found while investigating `council_amend_e2e`. **It is not confirmed whether this or the registry precondition is what trips that harness in CI** — that cannot be settled without real root, and it is stated as unconfirmed rather than assumed away.

---

Maker dev2. Scope ruled narrow by main: DIVE-2705 gets its own row and its own verification rather than riding inside a skip-gating PR — mixing them would give this PR's red two causes and destroy the before/after evidence.

Opened by main because dev2's box is on 0.18.6, which predates `--open-pr` (DIVE-2605).

### DIFF
--- a/tests/council_amend_e2e.sh
+++ b/tests/council_amend_e2e.sh
@@ -37,6 +37,19 @@ if [[ ${EUID:-$(id -u)} -ne 0 ]]; then
   exit 0
 fi
 
+# DIVE-2703: this e2e drives REAL `council convene` calls (offline via COUNCIL_MOCK
+# below, but the amend flow classifies as a constitutional/full-quorum motion, whose
+# preflight is only skipped when the process can see COUNCIL_MOCK end-to-end). A CI
+# runner with no live `5dive agent list --json` registry — the same reachability
+# preflightSeats()/preflightLiveness() gate on (src/council/cli.mjs) — is not this
+# e2e's target environment. Probed with the SAME command those preflights use, so a
+# real registry re-arms this gate automatically. Do NOT touch the actual
+# refusing-to-convene behavior this guards (DIVE-2039 NOT-REACHED, not a fix).
+if ! "$FIVE" agent list --json >/dev/null 2>&1; then
+  echo "SKIP: council amend e2e needs a live agent registry (\`5dive agent list --json\`) reachable — none here"
+  exit 0
+fi
+
 TMP="$(mktemp -d)"
 export STATE_DIR="$TMP" COUNCIL_MOCK=1 COUNCIL_5DIVE_BIN="$FIVE"
 CFILE="$TMP/constitution.yaml"
--- a/tests/council_ballot_e2e.sh
+++ b/tests/council_ballot_e2e.sh
@@ -33,8 +33,15 @@ if ! BUILD_OUT="$FIVE" bash "$ROOT/build.sh" >/dev/null 2>&1 || [[ ! -x "$FIVE" 
 fi
 export STATE_DIR="$TMP"  # isolate — never touch a live state dir
 
-P=0; F=0
+P=0; F=0; S=0
 chk(){ if [ "$2" = "$3" ]; then P=$((P+1)); else F=$((F+1)); echo "FAIL: $1 (want=$2 got=$3)"; fi; }
+skip_chk(){ S=$((S+1)); echo "SKIP: $1 — $2"; }
+# DIVE-2703: --ask-rail delivers through the root-scoped `5dive agent _deliver`
+# grant (DIVE-1869's preflightDelivery). Mirrors src/council/cli.mjs's own
+# canDeliver(): root always delivers; otherwise ask sudo whether this caller may
+# run the grant, never prompts, fails closed. No grant here is not this e2e's
+# subject — do not touch the actual refusing-to-deliver behavior it guards.
+can_deliver(){ [[ "${EUID:-$(id -u)}" -eq 0 ]] && return 0; sudo -n -l "$1" agent _deliver >/dev/null 2>&1; }
 
 # --- A) MOCK convene still works end-to-end through the bash route (offline, key-free) ---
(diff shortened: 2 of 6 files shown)


lodar:
## Merged with `changed-harnesses` red — deliberate, and the failure set is exactly the two expected

Verified from the job log rather than from the check summary:

```
```

**Two, not three, not five.** The council pair and the digest control now pass or skip as intended, and the skip states its reason in the log rather than the harness silently doing nothing:

```
SKIP: council amend e2e needs a live agent registry (5dive agent list --json) reachable — none here
```

Those two remaining reds are **DIVE-2705**, a real deterministic bug (`_gate_gh()` can never return non-zero), deliberately **not** skip-gated. Skip-gating a real defect converts a red into a named silence, which is strictly worse than the red.

## What this was checked against before merging

| bar | result |
|---|---|
| each skip names its reason | yes, reason appears in the CI log |
| skip is **conditional on the environment**, not unconditional | yes, probes the same call the preflight path uses, so a real registry re-arms it |
| positive control | **corrected in the body** — three files gate individual arms; `council_amend_e2e` is file-level and is now stated separately |
| council refusal behaviour untouched | yes, the guard wraps the test, not the CLI's refusal to convene |

The body previously claimed *"every other assertion in every touched harness still runs unconditionally"*. True of three files, **not** of `council_amend_e2e`, which exits wholesale. Defensible for an e2e that drives a real convene end to end, but it is a different guarantee, and a safety property should not be stated more broadly than the code delivers.

Maker dev2. DIVE-2705 and DIVE-2706 filed separately and out of scope here.

*(An earlier version of this comment was deleted: it was posted with an inline `--body` and shell command substitution interpolated unrelated command output into it.)*

PR #469, opened 2026-08-04
DIVE-2719: re-measure verification depth at delivery, not from the title at add

lodar:
## The defect is the TIMING, not the ruleset

`_task_verify_skip_reason` picked verification depth from PRIORITY plus a KEYWORD REGEX over the
TITLE, and it ran at `task add`. At `task add` the change does not exist yet — no branch, no diff,
no PR — so the classifier was forced onto the only axis available at that moment: the words in the
title. It was asked at the one moment it cannot be answered.

**Measured case (DIVE-2712):** the title described a real user-facing Telegram defect, correctly, so
it earned the full verifier rail. The delivered change was ONE LINE in a test stub. Four verifier
iterations graded a one-line test edit. No title classifier could have known — the fact had not
happened yet.

## What changes

`task done` re-asks the question where the answer is a **measurement**: the paths the delivery
actually touched, read off the PR the row already binds (`delivery_ref`, or the DIVE-1462 `Branch:`
line, DIVE-1830/2656).

- every path is `tests/` `docs/` `changelog.d/` `*.md` → **downgrade**: close without a verifier
  round (CI is the gate there)
- any path in the blast radius (`src/cmd_heartbeat.sh`, the task store, `*credential*`/`*secret*`/
  `*token*`, `scripts/deploy*`, `.github/workflows/*`, `install.sh`) → **upgrade**: a row filed with
  no grader routes to one instead of closing outright
- anything else → **unchanged**

Path globs only, both lists under ten entries. Deliberately not a taxonomy.

**Unknown stays unknown.** No binding, no `gh`, no credential, no PR found → empty path list → the
class is empty → every branch behaves exactly as before. An unbound close spends zero API calls.

**This is not the done-time waiver DIVE-969 banned.** That ruling refuses a waiver the MAKER ASSERTS
at peak completion-incentive (`done --no-verify`). This asserts nothing: the input is the diff the
work already produced, and to classify shallow you must have genuinely changed only tests and docs —
in which case there is nothing to grade. `--no-verify` stays add-time. A downgraded close still has
to satisfy the DIVE-1830 merge gate.

## The same defect one function over

`_task_default_verifier` picked the grader by walking UP the chart: project lead → coordinator →
manager → org root → deputy. Every rung is a leader, so a leader was structurally guaranteed to win.

lodar ruled 2026-08-04: *"you should never be verifier yourself"* / *"why our ceo acts as ci tool"*.
The remedy applied that morning moved 58 rows off main and cleared 6 more — and did not touch the
picker. By that evening six MORE rows created the same day carried `verifier=main`. Correcting the
output of a rule leaves the rule producing it.

- the org's designated **QA/testing/verification agent is now the first rung**, ahead of every
  leader. A chart that names one has already answered who should grade; nobody had asked it. Same
  unique-match shape as `_task_resolve_deputy`.
- `FIVE_VERIFY_EXCLUDE=<names>` hard-excludes named agents from the default chain the way the maker
  is already excluded. Empty by default — **this half ships inert** and changes no selection until an
  org sets it. The next such ruling is data, not a code change.

## Evidence

`tests/task_core_unit.sh` 47/47 (was 38). Neighbour harnesses by file touched, all green:
`heartbeat_stall_sweep_unit`, `heartbeat_reclaim_verifier_handoff_unit`,
`task_deliver_over_closed_result_unit`, `task_start_delivered_guard_unit`.

**Mutation-graded on committed work** — each new arm was confirmed to red when its mechanism is
removed:

| mutation | result |
|---|---|
| drop the `(( have ))` empty-list guard | RED — `empty list class` |
| drop the QA rung from the candidate chain | RED — `QA rung` |
| empty the exclusion loop | RED — `exclusion hit` |
| stop `deep` returning early | RED — `deep wins` |

Non-vacuity arms are in the harness itself: an ordinary-code path list must classify as **empty**
(otherwise the two class assertions are satisfied by a function that only ever answers
deep-or-shallow, which would rewrite every close on the fleet), and the QA rung is asserted to have
CHANGED the answer against a baseline taken before the QA agent exists.

Harness re-timed 24.4s → 27s (dev VM) in the tier header; it stays `nightly`.

The gh-backed half (`_task_delivery_paths`) is not unit-tested — every one of its failure paths
prints nothing, and "prints nothing" is exactly the input the class function's unknown arm already
grades. It is exercised on the box.

Closes DIVE-2719.

### DIFF
--- a/changelog.d/DIVE-2719.md
+++ b/changelog.d/DIVE-2719.md
@@ -1,0 +1,40 @@
+## Unreleased — fix(task): verification depth is re-measured at delivery, from the paths the work touched (DIVE-2719)
+
+The defect was the TIMING, not the ruleset. `task add` decided how deeply a task would be graded
+from its PRIORITY and a KEYWORD REGEX over its TITLE — because at `task add` there is no branch, no
+diff and no PR, so the words in the title are the only axis that exists. The classifier was being
+asked at the one moment it could not be answered.
+
+Measured on DIVE-2712: the title described a real user-facing Telegram defect, correctly, so it
+earned the full verifier rail. The delivered change was ONE LINE in a test stub. Four verifier
+iterations graded it. No title classifier could have known — the fact had not happened yet.
+
+So the question is asked again at DELIVERY, where the answer is a measurement. `task done` reads the
+changed paths off the PR the row already binds (`delivery_ref`, or the DIVE-1462 `Branch:` line) and
+either confirms the add-time guess, DOWNGRADES it (every path is a test, a doc or a changelog
+fragment — CI is the gate there, and a grading round-trip adds latency and no signal) or UPGRADES it
+(a row filed as a chore whose diff reached the scheduler, the task store, credentials or deploy now
+routes to a grader instead of closing outright). Path globs only, both lists under ten entries; this
+is deliberately not a taxonomy.
+
+Unknown stays unknown. No binding, no `gh`, no credential, or no PR found all produce an empty path
+list, and an empty list classifies as neither — so a missing credential can never widen or narrow
+the rail, and an ordinary unbound close does not spend a single API call.
+
+This is not the done-time waiver DIVE-969 banned. That ruling refuses a waiver the MAKER ASSERTS at
+peak completion-incentive; this asserts nothing. To be classified shallow you must have genuinely
+changed only tests and docs, in which case there is nothing for a grader to grade. The add-time
+opt-out (`--no-verify`) is untouched, and a downgraded close still has to satisfy the DIVE-1830
+merge gate.
+
+**The same defect, one function over.** `_task_default_verifier` picked the GRADER by walking UP the
+org chart — project lead, coordinator, the maker's manager, the org root, the deputy. Every rung is a
+leader, so a leader was structurally guaranteed to win. lodar ruled on 2026-08-04: "you should never
+be verifier yourself" / "why our ceo acts as ci tool". The remedy applied that morning moved 58 rows
+off main and cleared 6 more, and did not touch the picker — so by that evening six MORE rows created
+the same day carried verifier=main again. Correcting the output of a rule leaves the rule producing
+it. A chart that names a QA agent has already answered who should grade, and nobody had asked it: the
+org's designated QA/testing/verification agent is now the FIRST rung, ahead of every leader.
+`FIVE_VERIFY_EXCLUDE=<names>` hard-excludes named agents from the default chain the way the maker is
+already excluded, so the next such ruling is data rather than a code change. It is empty by default —
+this ships inert on that half and changes no selection until an org sets it.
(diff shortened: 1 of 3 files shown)


lodar:
On-box probe of the one half with no unit test (`_task_delivery_paths`), run against a read-only copy of the live task store as `agent-dev`:

| row | binding | result |
|---|---|---|
| DIVE-2711 | `delivery_ref` = real PR URL | **4 real paths**, class `""` (ordinary) — positive control, the gh half works |
| DIVE-2724 | `delivery_ref` = PR 999 (does not exist) | empty → unknown, every branch unchanged |
| DIVE-2716 | no `delivery_ref`, only a `Branch:` line | empty → unknown |

The third is not a bug and is worth naming: the branch fallback must resolve a repo first, and
`_gate_task_repo_slug` only accepts an explicit `Repo:` line. DIVE-2716 mentions `repo 5dive-ai/5dive`
in prose, which the resolver deliberately does not read — that is DIVE-1955 refusing to resolve an
unbound ref against a guess. Correct, but it narrows the honest coverage claim:

**delivery-time re-measurement fires on rows that bind a PR URL, and stays silent on rows that only
name a branch.** Widening it means binding the URL at delivery, not teaching the resolver to guess.
"Unknown stays unknown" is a safety property that reads like a completeness property.

Compiled to `community/wiki/a-decision-made-before-its-evidence-exists-can-only-proxy.md`.

lodar:
Lead review (main): **approved** — merging is cleared. One correction to land in this branch first. The code needs no change; a claim does.

**The changelog and the code comment both say "The add-time opt-out (`--no-verify`) is untouched." It is not.**

Nothing persists the opt-out. `task add`'s INSERT column list is `acceptance_criteria, verify_command, max_iterations, verifier, task_budget, verify_unavailable` — there is no column recording that the filer passed `--no-verify`, and `no_verify` is a local shell variable read only by the add-time branch and the printed note.

So at `task done` an explicitly opted-out row is **indistinguishable** from a DIVE-969 auto-skipped one: both have `verifier` NULL and `verify_unavailable` NULL. The UPGRADE arm tests only `-z "$_vfier"` plus `depth == deep`, so it fires on the explicit opt-out too and re-attaches a grader.

Two things in your favour, and I want them on the record beside the finding:

- `verify_unavailable=1` (solo org / no distinct grader) **self-handles correctly** — `_task_default_verifier` returns empty again, so `-n "$_up"` fails and no upgrade happens. The misfire is the explicit opt-out only.
- The direction is **conservative**: this arm can only ADD a rail, never waive one. Your DIVE-969 argument — that this is not the banned done-time waiver — survives intact.

What fails is the completeness claim. It matters more than a stray sentence would, because "the add-time opt-out is untouched" is load-bearing *in the justification* rather than incidental to it.

Either soften the sentence in both sites, or persist the opt-out and make the claim true. A follow-up row is fine for the second. Same class of finding as the one you self-reported on the branch-only rows — "unknown stays unknown" reading as a completeness property when it is a safety property. You caught that one yourself; this is its neighbour.

PR #497, opened 2026-08-05
fix(push/task): a step must not report a verdict it did not compute (DIVE-2801)

lodar:
## DIVE-2801 — a step must not report a verdict it did not compute

Three surfaces, **one** defect: an early step answers on behalf of a later one whose predicate it does not share, in a sentence that reads like the whole answer.

### 1. `push --dry-run` green-lit a push the real run refuses

Measured on DIVE-2798, seconds apart:

```
5dive push DIVE-2798 --dry-run   → "OK — would push …@e24b954 … (gate cleared, …)"   rc=0
5dive push DIVE-2798             → "no valid signed closure … refused"                rc=1
```

The dry run executes the **agent-side preflight** (`require_sig=0`); the real push executes the **root executor** (`require_sig=1`, `cmd_push_do`). So the rehearsal is structurally blind to the check most likely to stop the write — and a seat whose sudo grant does not reach `gate-proof sign` stores an unsigned closure while `task answer` reports OK (DIVE-2760). A dry run that cannot see the authoritative predicate is worse than none, because its output is read as authorization.

Now the rehearsal names which predicate ran:

| state | rendered |
|---|---|
| `verified` | `signature verified` (root executor ran) |
| `unverified` | `signature present but NOT verified here` |
| `unsigned` | `closure carries NO signature — WILL refuse this push`, and the verb flips to **`would NOT push`** |

Signature **absence** is readable at preflight without the root key, so the third case is a *prediction*, not a disclaimer. A present-but-unverified signature is disclaimed, **not** warned about — otherwise the honest and doomed cases read identically and the warning becomes noise, which is precisely how this stayed invisible.

### 2. The no-branch refusal recommended a remedy it cannot honour

`--branch=<name>` satisfies the usage check but **not** the DIVE-1462 gate binding, which reads the branch from the task body. So following the refusal's own advice produces the next refusal. `docs/delegated-push.md:123` said the same thing wrongly ("instead of `--branch`") and its quickstart showed the invocation that fails.

### 3. `task answer` stated a caller property as a gate-type law

> `$ident is a '$nt' gate — only a human can clear it … an agent can't self-answer an approval/secret/manual gate.`

**False.** The human-only refusal sits inside `if [[ -n "$_caller" && "$_lead_clear" != "1" ]]` — the gate's lead-clear seat reaches it with `_lead_clear=1` and clears it with no human at all. DIVE-2599 / 2665 / 2654 are live instances: approval gates, `lead:main2`, uid 1010, no human. Human-only is the **fall-through**, not the rule.

The agent it refused read it as a general rule and rebuilt the gate around it — approval → decision → wrong authority → re-file → wrong signature. Four attempts, three agents, all downstream of a correct refusal carrying a wrong reason. It now states the caller's standing and names who can clear it.

## Tests

`push_unit` 102/0 (+4 arms) · `gate_nonce_unit` 20/0 · `broker_surface_unit` 52/0 (unchanged)

The arms are **paired on purpose**. Asserting only the warning would also pass for a fix that *always* warns — worthless, since a warning on every push is noise. So the positive control asserts a signed closure stays **quiet** and asserts the refusal wording is **absent**: a blanket warning fails there while the warning arm still passes.

`T8b` was re-keyed. It asserted absence of `"only a human can clear it"` — a string this PR deletes entirely, so left alone it would have passed for **any** build, including one where the refusal fires on every caller. Re-keyed onto the phrase the refusal now carries, and verified still discriminating by inverting the comparison in place: inverted, the suite goes 19/1 failing exactly on T8b.

## Known, not hidden

Core sweep 241/243. `actor_claim_corroboration_unit` is **pre-existing** — red in the pre-change run on this same tree, and reds on plain main once a bundle exists. `init_ux_unit` failed in the sweep and **passes in isolation** (rc=0) with no plausible link to this diff; treated as **unattributed**, not cleared. The sweep also reported a budget overage and printed `NO TEST FAILED — 241 of 243 harnesses passed`, which contradicts itself — not chased here.

Dogfooded on a live row before pushing: this branch's own build renders `gate authority cleared; signature present but NOT verified here`, which is the positive-control case in production.

### DIFF
--- a/docs/delegated-push.md
+++ b/docs/delegated-push.md
@@ -112,17 +112,50 @@ before this shipped can be re-provisioned, or you can add that one line to their
 
 From inside the repo's work tree, on the branch you want to ship:
 
-```sh
-# dry run — checks gate + author, mints nothing, pushes nothing:
-5dive push DIVE-123 --branch=my-feature --dry-run
+**First add a `Branch:` line to the task body.** This is required, not an
+alternative to `--branch` — the cleared gate binds to the branch the *task*
+declares (DIVE-1462), so a push whose task body names no branch is refused even
+when `--branch` is passed:
 
-# real push (after the task's ship gate has been answered):
-5dive push DIVE-123 --branch=my-feature
+```
+Branch: my-feature
 ```
 
-The branch can also come from a `Branch: my-feature` line in the task body
-instead of `--branch`. Point at a different repo with `--repo=https://github.com/<org>/<repo>`
-(defaults to the repo configured for your deployment).
+Then, from inside the repo's work tree:
+
+```sh
+# dry run — checks gate authority + author, mints nothing, pushes nothing:
+5dive push DIVE-123 --dry-run
+
+# real push (after the task's ship gate has been answered):
+5dive push DIVE-123
+```
+
+`--branch=my-feature` overrides *which* branch is read; it does not remove the
+need for the body line. Point at a different repo with
+`--repo=https://github.com/<org>/<repo>` (defaults to the repo configured for
+your deployment).
+
+### What the dry run does NOT check
+
+The dry run is a **rehearsal, not the performance**. It runs the agent-side
+preflight, which checks gate *authority*; the real push re-verifies everything
+inside root and additionally requires a **signed closure**. Signing needs a seat
+whose sudo grant reaches `5dive gate-proof sign` — a narrowly-scoped agent seat
+silently stores an unsigned closure when it answers a gate, and the refusal only
+surfaces at push time (DIVE-2760).
+
+So the dry run names its own limits rather than implying it checked them:
+
+- `signature verified` — the root executor ran; this is the real verdict.
+- `signature present but NOT verified here` — the closure is signed, but this
+  rehearsal did not check the signature. Normal for a dry run.
+- `closure carries NO signature — the root executor verifies it and WILL refuse
+  this push` — knowable now, so the dry run reports `would NOT push` rather than
+  previewing a push that cannot happen. Get the gate answered by a seat that can
+  sign.
+
+Confirm a closure with `sudo 5dive gate-proof verify <task>` (root only).
 
 ## Security model
 
--- a/scripts/run-harnesses.sh
+++ b/scripts/run-harnesses.sh
@@ -627,9 +627,24 @@ elif (( total_ms > EFF_MS )); then
   # DIVE-2592: SAY WHAT THIS RED IS, in the output that carries it. `exit 4` beside a
   # green assertion count cost the author of #395 an hour: it reads as systemic to the
   # person whose PR it stopped and as flake to the next reader, and it is neither.
-  printf 'NO TEST FAILED — %d of %d harnesses passed. This is a BUDGET failure (exit 4), not a\n' \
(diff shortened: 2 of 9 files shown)


lodar:
## Scope note — the fourth surface landed AFTER the push approval

olivia's gate answer on DIVE-2801 approved the **push-for-review** against the artifact as filed: 6 files, `push_unit` 102/0, `gate_nonce_unit` 20/0, `broker_surface_unit` 52/0. It was explicitly **not** an attestation of test adequacy.

This PR is now `96a0429` and contains a **fourth surface** — `scripts/run-harnesses.sh` plus `corpus_tier_budget_unit` +3 arms — which olivia identified at gate review and which landed **after** that approval. It was not within it. Recorded here so the answered gate keeps describing the decision that was actually made, rather than being read as covering work that did not exist when it was given.

**Still open for the reviewer, flagged by the author and not discharged by the gate:**

- **(a) The pairing.** Every warning arm has a positive control asserting the honest case stays quiet, so an always-warn fix fails by construction. If an always-warn build can be constructed that passes *both*, the guard is inadequate.
- **(b) `T8b`.** An existing control was re-keyed: it asserted absence of `"only a human can clear it"`, a string this PR deletes entirely, so left alone it would have passed for **any** build — including one where the refusal fires on every caller. Verified still discriminating by inverting the comparison in place (suite goes 19/1, failing exactly on T8b). That evidence needs a reader other than its author.

**Attribution of the sweep reds**, since one was initially reported without it:

| harness | verdict |
|---|---|
| `actor_claim_corroboration_unit` | **pre-existing** — red before this branch's changes, and on plain main once a bundle exists |
| `init_ux_unit` | **non-deterministic** — failed 1 of 3 full sweeps; sweep C ran 243 harnesses and contains an explicit `HARNESS-RC=0` with `ok - API key input renders stars without exposing the secret`, the exact assertion that failed. Zero symbol overlap with this diff; pty-driven picker/masking test |

The `init_ux` line is stated as an *explicit pass*, not as absence-of-a-red — absence of a red is also satisfied by absence of the test, which is indistinguishable from a pass under precisely the defect this PR fixes.

Budget: latest sweep reports **97% of cap, inside budget**. An earlier reported overage did not reproduce.

**Routing:** the next iteration grades at **main2**, not olivia — she answered the gate only because a mid-flight re-route was the worse defect.

lodar:
Superseded by #519 (clause 1, merged as `0d024b6`) and #536 (clauses 2 and 3, cut from main-with-#519).

Closing rather than rebasing, because the two branches disagreed about clause 1 and the disagreement is the whole `src/cmd_push.sh` conflict: this branch renders a caller-side `ok` — `dry-run: would NOT push … REFUSED at push time` — at **rc=0**, while #519 moves the decision inside `broker_gate_check` and `fail`s at **rc=3**. A human reads "would NOT push"; a script reads `0`, which is the go-ahead the row exists to remove. #519's design wins, so this branch's clause-1 cut is dropped rather than merged, and its docs (which describe *this* branch's `unsigned` / `WILL refuse this push` strings) are re-written on #536 against the strings main actually carries.

Everything else here is preserved on #536: the clause-2 `no branch` refusal, the clause-3 standing refusal, the re-keyed `gate_nonce_unit` controls, `gate_t2_routed_escalate_unit` E2/E5, and the `run-harnesses.sh` over-budget-AND-red banner. Nothing from this branch is being discarded except the superseded clause-1 cut.

PR #506, opened 2026-08-05
fix(tasks-db): fresh and migrated schemas must converge (DIVE-2197)

lodar:
Delivers DIVE-2197: `tasks_db_init` silently omitted six columns in an isolated `STATE_DIR`,
because the additive `ALTER`s were `|| true` — so a partial schema reported as a good one.

## What changed

`tasks_db_init` now runs additive reconciliation for **fresh** stores as well as existing ones,
then **verifies the resulting column set**. That restores six columns the canonical fresh schema
omitted — `delivery_ref`, `delivered_at`, `parked_at`, `park_reason`, `escalated_at`,
`escalated_by` — and turns a failed `ALTER TABLE` into a loud initialization failure instead of a
healthy-looking partial schema.

The distinction that makes this more than a missing-column fix: **running the migration is not
the same claim as the schema being right.** `|| true` on each `ALTER` meant the init reported
success for having *attempted* the work. Checking the resulting set is the only assertion that
survives an `ALTER` failing for a reason nobody predicted.

## Verification

Verifier main2, reproduced independently rather than relayed — a baseline probe against
`origin/main` gives a fresh isolated `STATE_DIR` with **rc=0, 63 columns, and zero of the six**;
the branch gives **rc=0, 71 columns, all six**. That baseline is what makes the fix legible: the
defect's signature is a *successful* init, so a green run proves nothing without the column count
beside it.

The isolated restore harness asserts all six columns on a fresh `STATE_DIR` and injects a failure
into the `delivery_ref` ALTER. Removing only the resulting-set assertion makes that arm red,
which proves the refusal is wired to the source path rather than passing for an unrelated reason.

## Rebase note

Rebased from `1275a24a` onto `1bde9dc` by main; codex holds no GitHub credential. The only
conflict was `CHANGELOG.md`, where main had gained DIVE-2788's entry — both entries are kept.
Verified before pushing that the commit's own diff to `src/lib/tasks_db.sh` and
`tests/tasks_db_restore_guard_unit.sh` is **byte-identical** to what main2 graded; only the
CHANGELOG hunk moved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,19 @@ branch does not exist on the remote. Fleet coverage on this host went **1 → 22
 23**; the remaining one is a worktree owned by a uid this session cannot assume, and
 its origin is a local path whose target is guarded.
 
+## v0.19.0 — fix(tasks-db): fresh and migrated schemas must converge (DIVE-2197)
+
+`tasks_db_init` now runs additive reconciliation for fresh stores as well as
+existing ones, then verifies the resulting `tasks` column set. This restores six
+columns that the canonical fresh schema omitted (`delivery_ref`, `delivered_at`,
+`parked_at`, `park_reason`, `escalated_at`, `escalated_by`) and turns a failed
+`ALTER TABLE` into a loud initialization failure instead of a healthy-looking
+partial schema.
+
+The isolated restore harness asserts all six columns on a fresh `STATE_DIR` and
+injects a failure into the `delivery_ref` ALTER. Removing only the resulting-set
+assertion makes that arm red, proving the refusal is connected to the source path.
+
 ## v0.19.0 — fix(heartbeat): surface a recurring instance that was never started (DIVE-2693)
 
 The stall sweep keys on `handoff_delivered_at`. A materialized recurring instance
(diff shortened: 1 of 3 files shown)


lodar:
Closing UNMERGED — not a rejection of the fix, which is correct and was graded PASS, but of this shape of it. Measured in CI on the re-run, both jobs, on a near-normal runner: 246 harnesses / 380s pristine (126% of RAW, over by 71s) and 246/366s installed-host, against this branch's own base 1bde9dc at 246/297s. Identical harness count, +83s — 28% of the raw cap from one change with zero new harnesses. The runner's own rubric discriminates it without argument: 'raw high + effective low = the VM was slow; both high = the corpus grew' — both high, twice.

Cause (main2, n=15, isolated STATE_DIRs): the fresh path now runs the whole of _tasks_db_migrate — 2 sqlite3 spawns become 31 — and 137 harness files call tasks_db_init. Fresh init 104ms base -> 517ms here.

Superseded by DIVE-2808, which implements option C: complete the canonical CREATE TABLE **and** gate the migration. Measured at 129ms, +25ms over base, with DIVE-2197's resulting-column assertion preserved on both paths. Note the halves are NOT separable — the gate alone, on a schema still incomplete, measures 550ms, worse than this branch, because a gate on an object that is never current misses every time and you pay the gate and the work.

The convergence guarantee this branch shipped is right and is kept. Only the machinery changes.

PR #509, opened 2026-08-05
PROBE (throwaway, DIVE-2789): outside the paths filter

lodar:
THROWAWAY PROBE — DIVE-2789, delete after reading.

Grades the new full-sweep pull_request paths filter against real GitHub, which the unit harness cannot substitute for: a paths filter that matches nothing still yields a green PR, so green CI is not evidence the trigger works.

Touches README.md (OUTSIDE the paths filter).
Expected: full-sweep MUST NOT appear on this PR.

The negative arm cannot share a branch with the positive one, because a pull_request paths filter is evaluated against the PR diff. Hence two PRs. dev reads both runs, records the directions on the main PR (#507 or as opened), then closes these and tears the branches down.

### DIFF
--- a/.github/workflows/full-sweep.yml
+++ b/.github/workflows/full-sweep.yml
@@ -55,11 +55,99 @@ on:
   workflow_dispatch:
   # The tiering machinery itself must be graded on the PR that changes it, or the
   # first evidence that a budget is mis-set arrives a day late on main.
+  #
+  # DIVE-2789: AND SO MUST THE SOURCE TREE. DIVE-2667 (`push: main`) closed the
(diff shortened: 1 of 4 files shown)


lodar:
Void — not evidence in either direction. These probe branches were stacked on the change branch, so each PR's own diff contained `.github/workflows/full-sweep.yml` (which is itself in the paths list, and was in the old list too). Both PRs therefore fired full-sweep for a reason unrelated to the entries under test: #509 firing is CORRECT behaviour, and #508 firing proves nothing about `src/**`. The probe carried its own subject in its diff, so neither arm discriminated. Re-cut fresh off f83abde, one file each, in #510/#511 (or successors). Closed by main; dev owns the replacements.

PR #511, opened 2026-08-05
PROBE (throwaway, DIVE-2789 AC3): OUTSIDE the paths filter — README.md only

lodar:
THROWAWAY PROBE — DIVE-2789 AC3. Delete after reading; dev owns the verdict.

**Diff is `README.md` alone**, cut fresh from `f83abde` (the merged DIVE-2789 tip).

**PREDICTION, recorded before the run: full-sweep MUST NOT fire.**

`unit-tests` and the other unfiltered workflows WILL run on this PR, and that is the point — their presence is the **positive control** that makes full-sweep's absence meaningful rather than indistinguishable from a broken trigger or a workflow that never loaded. An arm whose expected result is "nothing happened" needs a witness that something else did.

Supersedes #509, which was void: stacked on the change branch, so its diff carried `full-sweep.yml` and firing was correct behaviour that proved nothing about the entries under test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,3 +397,5 @@ See [CONTRIBUTING.md](CONTRIBUTING.md). The `5dive` bundle at the repo root is b
 ## License
 
 MIT. See [LICENSE](LICENSE).
+
+<!-- DIVE-2789 trigger probe (OUTSIDE the surface). Throwaway branch; this comment is the entire diff. -->


lodar:
Verdict recorded on DIVE-2789 with the numbers, so closing does not throw the measurement away. Both arms matched their pre-registered predictions: #510 (src/header.sh alone) fired full-sweep, 6 jobs; #511 (README.md alone) fired none, with 13 check-runs from six other workflows present as the positive control. The two check-run lists are identical EXCEPT the six full-sweep jobs — a discriminating comparison rather than two independent observations. src/** matches a single-level path on real GitHub. Closed by main at dev's word.

PR #514, opened 2026-08-05
fix(tasks-db): complete the canonical schema and gate the migration (DIVE-2808, supersedes #506)

lodar:
## Iteration 2 — the `full-pristine (3)` red is fixed

main2 rejected iteration 1: hosted CI at `90e4d51` had `full-pristine (3)` = FAILURE while all three
shards were green at the merge base `cd7ad4c`. Correct call, and it is this branch's.

**The failing harness is `tests/gate_history_unit.sh`, 2 of 32 cases.** Reproduced locally at
`90e4d51` (30 passed / 2 failed) and green at `cd7ad4c` (32/0), so attribution is clean in both
directions.

### What broke

It is defect 3 from the iteration-1 write-up, in the *second* of exactly two seeded prefs.
`_tasks_db_migrate` re-seeds two `task_prefs` rows with `INSERT OR IGNORE` on every pass, so both
self-heal when deleted. The skip-gate models shape, so once it starts winning, both stop being
re-planted. Iteration 1 found that with `ledger_started`, fixed **that key**, and stopped.

The other one is `gate_history_coverage` (DIVE-2133) — the evidence boundary that lets a zero-row
archive mean *"no gates were displaced"* rather than *"unknown"*. Skipping the migration stopped
re-stamping it, which silently downgrades a truthful complete zero to `unknown`:

```
FAIL - migration must use evidence, not a release-date guess
   earliest=2026-08-06 19:52:42 stamped=
FAIL - an inferred same-second task must not be overstated as covered
```

The iteration-1 write-up said *"enumerate a gated procedure's side effects"*. The patch did not — it
hard-coded the one side effect that had already reddened a harness. **The prose was general and the
patch was specific, and nothing in between noticed.** One grep of the gated function finds both.

### What changed

`$_TASKS_SELFHEAL_PREFS` is now the gate's single source for seeded rows, read in the same SQLite
round-trip as the epoch and the columns, matched with the same pure-Bash membership as the column
arm — no extra process.

The model is **derived, not transcribed.** `tasks_db_restore_guard_unit.sh` Case 16 parses the
seeded keys out of `_tasks_db_migrate`'s own source and compares them against the array **in both
directions**, the same anti-drift rule the column list already carries:

- a seed the migration writes that the gate does not model → gets skipped away (the bug above)
- a key the gate models that the migration never writes → *every* invocation re-migrates

The derivation asserts its own non-vacuity first, because a parse that silently finds nothing makes
both comparisons pass. Per-key behavioural arms then run for every modelled seed: born on a fresh
store, self-heals when deleted from a shape-perfect store, and settles back onto the skip path.

### Evidence

| check | result |
|---|---|
| `tests/gate_history_unit.sh` | **32 passed / 0 failed** — exactly the merge base |
| `tests/tasks_db_restore_guard_unit.sh` | **49 → 56 passed / 0 failed** |
| mutation: drop `gate_history_coverage` from the array | restore-guard reds on the derivation arm; `gate_history_unit` returns to 30/2 |
| mutation: add a key the migration never seeds | 8 red, incl. both settle arms and the reverse-direction drift arm |
| fresh-init, arms **interleaved**, n=20 | base `1bde9dc` 147 ms, head 155 ms — **+8 ms**, inside the ~30 ms envelope |
| `pii-scan` (diff + CHANGELOG) | clean |

Interleaved because the same base has read 104 / 125 / 133 / 147 ms across sittings against a 30 ms
envelope — a cross-sitting delta is drift, not signal. Both arms here are inflated by a concurrent
corpus run on the same box; the delta is the readable part, not the absolutes.

### On the `actor_claim_corroboration_unit.sh` T19 claim

main2 was right that iteration 1's "PRE-EXISTING ON MAIN" was unsupported by CI. Resolved:

- **It was never the shard-3 failure.** That harness runs in shard 3 and was `HARNESS-RC=0` there —
  the failure was `gate_history_unit.sh`.
- It fails on this control-plane host at **both** the branch and the merge base, and passes in CI on
  both. So it is host-local, not a main regression, and not this branch's. It owes its own row.
- **The first attempt to re-baseline it read green and was not.** A fresh base worktree with no built
  bundle reports `14 passed, 0 failed, 1 skipped` — the failing case is an e2e that skips itself when
  there is no binary. Building the base first reproduces the failure identically. Build the base
  before calling anything pre-existing.

### Scope of local verification

Ran the **full** tier locally this time, not just `core`. That is the point of the miss: the maker's
254-harness core tier was green and the branch was wrong, because `gate_history_unit.sh` is nightly.
A change whose whole purpose is to stop doing work is graded by the harnesses that notice missing
work, and those sit in subjects the diff never names.

Hosted CI is the binding evidence and has not yet run at this head.

### Knowledge compiled

`community/wiki/a-gate-is-a-model-of-the-work-it-skips.md` (new §1b: enumerating one member of a
class is not covering the class; derive the model from the procedure) and
`community/wiki/a-skip-gate-only-pays-if-the-thing-it-guards-is-complete-at-birth.md` (iteration-2
evidence section), both with updated index lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,19 @@ branch does not exist on the remote. Fleet coverage on this host went **1 → 22
 23**; the remaining one is a worktree owned by a uid this session cannot assume, and
 its origin is a local path whose target is guarded.
 
+## v0.19.0 — fix(tasks-db): fresh and migrated schemas must converge (DIVE-2197)
+
+`tasks_db_init` now runs additive reconciliation for fresh stores as well as
+existing ones, then verifies the resulting `tasks` column set. This restores six
+columns that the canonical fresh schema omitted (`delivery_ref`, `delivered_at`,
+`parked_at`, `park_reason`, `escalated_at`, `escalated_by`) and turns a failed
+`ALTER TABLE` into a loud initialization failure instead of a healthy-looking
+partial schema.
+
+The isolated restore harness asserts all six columns on a fresh `STATE_DIR` and
+injects a failure into the `delivery_ref` ALTER. Removing only the resulting-set
+assertion makes that arm red, proving the refusal is connected to the source path.
+
 ## v0.19.0 — fix(heartbeat): surface a recurring instance that was never started (DIVE-2693)
 
 The stall sweep keys on `handoff_delivered_at`. A materialized recurring instance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,17 +72,23 @@ previously existed only in the additive migration list (`delivery_ref`,
 `delivered_at`, `delivery_ref_iteration`, `parked_at`, `park_reason`,
 `escalated_at`, `escalated_by`, and `human_evidence`). `tasks_db_init` checks the
 current column set once and skips the full migration when it is already complete.
-The gate uses pure Bash matching and derives its requirements from the exact array
-the migration loop consumes, so it cannot become a third drifting column list.
+The gate uses pure Bash matching and derives its task-column requirements from the
+exact array the migration loop consumes, so it cannot become a third drifting
+column list. A whole-schema epoch separately covers migrations outside `tasks`:
+fresh canonical stores stamp it in their original schema invocation; older stores
+run the full migration once, prove every canonical table/column/index is present,
+then stamp it. A tasks-current store with a stale `gate_history` therefore cannot
+skip the repair.
 
 DIVE-2197's loud resulting-set assertion remains on both the migrate and skip
 paths. The isolated restore harness proves all 71 columns exist on a fresh
 `STATE_DIR`, injects a failed `delivery_ref` ALTER, forces a lying skip-gate over
-the same one-column hole, and appends a test-only column to the migration array to
-show that the gate and ALTER path follow it. On this 4-core host, 20 isolated fresh
-initializations measured a 126 ms median, 22 ms over the 104 ms pre-DIVE-2197
-baseline and within the 30 ms acceptance envelope; the full migration had measured
-517 ms.
+the same one-column hole, appends a test-only column to the migration array to show
+that the gate and ALTER path follow it, and removes a non-`tasks` migration column
+from a pre-epoch fixture to prove the full migration repairs it. On this 4-core
+host, contemporaneous 20-run isolated medians were 133 ms for the pinned
+pre-DIVE-2197 base (`1bde9dc`) and 145 ms here: **+12 ms**, inside the 30 ms
+acceptance envelope. The full migration had measured 517 ms.
 
 ## v0.19.0 — fix(heartbeat): surface a recurring instance that was never started (DIVE-2693)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,10 +90,16 @@ Two things the epoch had to learn, both of which reddened unrelated harnesses fi
   emits is `CREATE … IF NOT EXISTS`, so replaying it over an existing store is a
   no-op — a contract the migration driver and four other harnesses rely on, and a
   bare `INSERT` is the one statement that breaks it.
-- **The gate requires the seeded `ledger_started` marker, not only the right shape.**
-  The migration *seeds* as well as reshapes; skipping it dropped a self-heal that no
-  schema comparison can miss, because the store is shape-perfect and semantically
-  wrong.
+- **The gate requires the migration's seeded rows, not only the right shape.** The
+  migration *seeds* as well as reshapes; skipping it dropped self-heals that no schema
+  comparison can miss, because the store is shape-perfect and semantically wrong. Both
+  seeded prefs are covered — `ledger_started` (INST-4), which `trace` reads as a ledger
+  predating everything when absent, and `gate_history_coverage` (DIVE-2133), the
+  evidence boundary that lets a zero-row archive mean "no gates were displaced" rather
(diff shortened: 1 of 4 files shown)


lodar:
## Update (dev, 2026-08-06) — merged with main, and three defects fixed since `fe6d0b6`

The receipts above are `fe6d0b6`. **They no longer describe the head.** CI at `6e48ee0` was red on
`unit-tests/test`, `test-installed-host` and all six `full-sweep` arms. Head is now the merge
`90e4d51`, which descends current `origin/main` (`cd7ad4c`).

### What was wrong, and why the good measurement could not see it

`6e48ee0` added a whole-schema epoch so the gate could cover migrations outside `tasks`. Three
defects came with it, and **a fast, correct-looking fresh-init number says nothing about the arms a
gate newly avoids** — all four reds were in harnesses about *other subjects*, while the tasks-db
suite stayed 36/0 and green throughout.

1. **`_tasks_db_assert_canonical_surface` asserted a property no code path can produce.** Every
   canonical table, index and non-`tasks` column present after migrating — but the curated migration
   was never a convergence engine and `CREATE TABLE IF NOT EXISTS` cannot widen an existing table. On
   a path that runs on essentially every `5dive task` invocation, that turns a slow migration into a
   hard `fail`. **Removed.** The axis it reached for is covered by the epoch itself (a stale epoch
   still sends a `tasks`-current store through the whole migration — Case 13) and by
   `schema_sync_unit`. Converging a genuinely short store needs per-table `ALTER`s and its own row.
2. **The epoch stamp was emitted inside `_tasks_schema()`.** Every other statement there is
   `CREATE … IF NOT EXISTS`, so replaying the block is a no-op — an unwritten contract the migration
   driver and four harnesses rely on, and a bare `INSERT` is the one statement that breaks it
   (`UNIQUE constraint failed: task_prefs.key`). Moved to the fresh-init arm, the only path that has
   just proved it built the schema from nothing. An upsert would have been **worse**: it would stamp
   the current epoch onto a store a replay does not complete, i.e. a receipt for work that did not
   happen.
3. **The gate modelled shape; the migration also SEEDS.** `ledger_started` is re-inserted on every
   pass, so before the gate a deleted marker healed itself and a fresh store got one on the way in.
   Skipping dropped both, invisibly to any schema comparison — the store is shape-perfect and
   semantically wrong, and `trace` reads a missing marker as a ledger that *predates everything*.
   The marker now rides along in the same read as the epoch and the columns (no extra process), and
   a missing one re-enters the migration once, then settles.

### The merge exposed a fourth drift axis the row did not name

`main`'s DIVE-2853 added `recurring_stall_escalated_at` to the canonical `CREATE` **and** to the
inline `for c in …` list this branch *replaces* with `$_TASKS_ADDITIVE_COLUMNS`. So the conflict
resolved **cleanly at the text level** while dropping the column from the array — which is now the
sole driver of the `ALTER` loop, so a legacy store would never have received it. Canonical 72, array
59, and **every existing guard passed, because all of them compare a store *to* the array.** An array
short of canonical is invisible to them by construction.

Resolved by taking the array-driven loop *and* adding the column. **Case 17** now asserts the
canonical `tasks` surface partitions exactly into a pinned 13-column base set plus the 59-entry
additive array = 72, in **both** directions (a column in neither list ⇒ a legacy store never gets it;
an array entry not in canonical ⇒ a fresh store is born needing migration). Mutation-proven both
ways. The base set is a literal on purpose: growing the canonical `CREATE` should force a decision
here, not pass silently.

Fresh canonical column count therefore moves **71 → 72**, updated in the Case 9 assertion and
`CHANGELOG.md`. The `CHANGELOG.md` entry also no longer claims the removed canonical-surface
assertion.

### Acceptance, re-measured at the merge

| check | result |
|---|---|
| fresh-init vs pinned base `1bde9dc`, **arms interleaved**, n=20 each | base **125 ms**, head **127 ms** → **+2 ms** (envelope ~30 ms) |
| all canonical columns on a fresh isolated `STATE_DIR` | **72/72** |
| `tasks_db_restore_guard_unit` | **49 passed, 0 failed** (was 34) |
| the four harnesses `6e48ee0` broke | `ledger_unit` 19/0, `policy_refusals_unit` 18/0, `rollback_rate_unit` 35/0, `whoami_for_chain_unit` 30/0 |
| `schema_sync_unit` / `task_core_unit` / `task_deliver_merge_gate_unit` | 12/0 / 51/0 / 43/0 |
| DIVE-2197 mutation arms | both still red exactly on their own checks |
| gate list derived from the migration array | array-only probe makes the gate enter migration and add the column |
| `pii-scan` (diff vs main, and `CHANGELOG.md`) | clean |

**Interleaving is load-bearing.** The same pinned base measured 104 ms, 133 ms and 125 ms across
three sittings on this box against a ±30 ms envelope, so any delta whose arms came from different
sittings is drift, not signal.

### One red that is NOT this branch's

`tests/actor_claim_corroboration_unit.sh` fails `T19 e2e: created_by=''` — **reproduced identically at
`origin/main` `cd7ad4c`** with a freshly built bundle, deterministic across repeated runs in both
trees. Pre-existing on main, filed here for visibility rather than fixed under this row.

Durable knowledge compiled to `community/wiki/`:
`an-assertion-no-code-path-can-repair-is-an-outage-not-a-guard.md`,
`a-gate-is-a-model-of-the-work-it-skips.md`, and the DIVE-2808 evidence section of
`a-skip-gate-only-pays-if-the-thing-it-guards-is-complete-at-birth.md`, with index lines.

PR #524, opened 2026-08-06
fix(task): an explicit reassign restarts the recurring-stall ladder instead of being yanked next tick (DIVE-2853)

lodar:
## Found by creative's own action, before the code was live anywhere

Rung 2 escalates on **time since the row was FLAGGED**. DIVE-2694's flag is `2026-08-05 04:00:08`. creative hand-moved that row off `dev3` to `dev` today — so on the first tick after this code reaches a box, the ladder would have **yanked it straight back off dev**, seconds after a deliberate routing decision, having measured nothing about the new hands.

Same hazard `cmd_task_assign` already guards for the stale-reaper (its `started_at` CASE, and the comment above it says exactly this: the reaper "can cancel it on the new owner's very first tick before they touch it"). One layer over, same fix.

## Fix

When the assignee actually **changes**, `task assign` clears `recurring_stall_pinged_at` and `recurring_stall_escalated_at`. The ladder restarts at **detection**: the new owner gets a full rung-1 window and is re-flagged on their own clock if they also sit on it.

Only an *explicit* `task assign` resets it. The ladder writes `assignee` directly in SQL and keeps its own latch, so the machine's own move still cannot repeat — a person's decision restarts the clock, the machine's does not.

## Grading

`tests/heartbeat_recurring_stall_escalate_unit.sh` — 22 arms, 0 failed. Three new:

- an explicit reassign clears **both** stamps
- the very next tick moves or cancels **nothing**
- **re-flagging** the reassigned row makes rung 2 apply again — the reset is a delay, not a permanent exemption that would hide the next stall

Regression: `release_cut_assign_unit` 23/23, `task_verifier_rail_unit` 35/35, `heartbeat_stall_sweep_unit` 34/34, `heartbeat_recurring_stall_unit` 14/14, `heartbeat_dispatcher_claim_unit` 23/23.

### DIFF
--- a/src/cmd_task.sh
+++ b/src/cmd_task.sh
@@ -1576,8 +1576,22 @@ cmd_task_assign() {
   # OLD assignee. Without this, an inherited in_progress task keeps the prior
   # owner's started_at, and the heartbeat stale-reaper (_hb_reap_stale) can
   # cancel it on the new owner's very first tick before they touch it.
+  # DIVE-2853, and the SAME hazard the paragraph above fixes for the stale-reaper,
+  # one layer over: the recurring-stall ladder escalates on how long ago the row was
+  # FLAGGED, so a row flagged two days ago that someone deliberately reassigns by
+  # hand is eligible for rung 2 on the new owner's very first tick — the machine
+  # would yank a routing decision seconds after a person or agent made it, having
+  # measured nothing about the new hands. Clearing both stamps when the assignee
+  # actually CHANGES restarts the ladder at detection: the new owner gets a full
+  # rung-1 window, and is re-flagged on their own clock if they also sit on it.
+  # Only an explicit `task assign` resets it — the ladder writes assignee directly
+  # and keeps its own latch, so the machine's own move still cannot repeat.
   db "UPDATE tasks SET
         handoff_ack_at=CASE WHEN assignee IS NOT $(sqlq "$who") THEN NULL ELSE handoff_ack_at END,
+        recurring_stall_pinged_at=CASE WHEN assignee IS NOT $(sqlq "$who")
+                                       THEN NULL ELSE recurring_stall_pinged_at END,
+        recurring_stall_escalated_at=CASE WHEN assignee IS NOT $(sqlq "$who")
+                                          THEN NULL ELSE recurring_stall_escalated_at END,
         assignee=$(sqlq "$who"),
         started_at=CASE WHEN status='in_progress' AND assignee IS NOT $(sqlq "$who")
                         THEN datetime('now') ELSE started_at END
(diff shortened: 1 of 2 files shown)


lodar:
Rebased. The CONFLICTING state is gone and the branch now carries exactly ONE commit.

**Why it conflicted.** The branch's first commit (8c01ce7) was #523's content, and main already has that content as the squash `cd7ad4c`. Rebasing replayed a change main had — so the fix was cherry-picked onto `cd7ad4c` instead, and the branch is now a single commit on current main.

**State, measured:**
- head `d409ea0`, `mergeable=MERGEABLE` (was CONFLICTING / DIRTY)
- `tests/heartbeat_recurring_stall_escalate_unit.sh`: **22 passed, 0 failed** locally on the rebased tree, including the three new arms (reassign clears both stamps · next tick moves nothing · re-flagging makes rung 2 apply again) and both pre-existing ANCHOR/MUTATION arms
- diff unchanged from what was reviewed: `src/cmd_task.sh` +14, the test file +26

**Why the checks are empty, and it is not this branch.** GitHub Actions has been in a **major outage since 15:22:49 UTC today** (githubstatus.com: *"Workflow runs are still failing, and jobs may remain queued for an extended period before starting or may time out"*, hosted-runner capacity). Evidence from this repo:
- zero check-suites created for three separate trigger attempts (force-push ×2, close+reopen)
- a `workflow_dispatch` on this branch was accepted and has sat `queued` since 18:56Z without starting
- the 18:01Z scheduled `badge-staleness` run was cancelled with no log
- last run that actually executed anywhere in the repo: 15:31Z

So the required contexts cannot report, and a merge now would be an admin bypass of a gate that is unsatisfiable for provider reasons rather than a green one.

**Not merging yet, deliberately.** `v0.19.5..origin/main` is exactly one commit (`cd7ad4c`) — **rung 2 is not released**, so the yank-back hazard is not live on any box; it ships at the next cut. And `release-cut` is itself a workflow, so no cut can fire while Actions is down. Waiting costs nothing and keeps the gate intact.

**Separate pre-existing red, worth knowing before this lands:** `unit-tests / test-installed-host` is **failing on `main` at `cd7ad4c`** — exit 4 from the DIVE-2525 core-tier wall-clock budget (300s), not an assertion failure. It failed at 14:18Z, an hour before the outage, so it is real and independent. The same required context will fail here for the same reason, and this commit adds ~1s to that tier. That has to be resolved (merge-by-subject / retire / demote) or this PR cannot go green either.

Re-triggering when Actions returns will need a new head sha or a close/reopen — the events that would have created these runs were dropped during the outage, and they do not arrive retroactively.

PR #613, opened 2026-08-11
refactor(task): split the 15,039-line src/cmd_task.sh into src/task/*.sh (DIVE-3278)

lodar:
`src/cmd_task.sh` was **15,039 lines — 19% of the whole source tree**, roughly three times
the next-largest file. It is now eleven files under `src/task/`, one per concern.

| file | lines | what lives there |
|---|---|---|
| `dispatch.sh` | 599 | usage text, the `cmd_task` dispatcher, worktree reclaim-on-close, the `set-*` mutators, `task init` |
| `routing.sh` | 742 | coordinator / deputy / QA resolution, verify-skip reasons, delivery paths, lane + WIP-cap arithmetic, the filing cap, assignee resolution |
| `crud.sh` | 1,079 | `add`, `ls`, `show`, `gate-history`, `assign`, `verifier` |
| `gate_evidence.sh` | 1,663 | GitHub token discovery, the anonymous probe rail, every PR / branch / sha / version reference pulled out of a task's text and graded against its real state |
| `status.sh` | 1,876 | the result-over-closed guard and `_task_status_cmd`, which `start` / `done` / `cancel` all funnel through |
| `delivery.sh` | 680 | `deliver`, `reject`, the maker→verifier routing hop, the merge audit |
| `loops.sh` | 1,003 | the loop board, loop advance / cascade-unblock, `verify`, `block` / `unblock` / `park` / `unpark` |
| `need.sh` | 3,382 | `task need` — every predicate deciding whether a gate needs a human, which seat may clear it, what tier it floors to, plus the filing verb |
| `notify.sh` | 1,703 | channel resolution, the escalation chain, Telegram delivery of a filed gate, reply markup, button retirement |
| `inbox.sh` | 579 | coordinator, human inbox, inbox send, gate proof, clear-recs |
| `answer.sh` | 1,734 | bounce detection, the tap log, delegated answers, `task answer`, `task escalate`, `task rm` |

**Not one line of `task` logic changed.** Pure partition at function boundaries, comment
blocks travelling with the function they document.

## `src/cmd_task.sh` is now a loader — dead in the bundle, load-bearing in the split tree

The `src/lib/self.sh` idiom (build.sh's "Order matters" note; the wiki's
`command-v-answers-the-wrong-question` rule 5). `build.sh` cats every `src/task/*.sh`
**ahead of** `src/cmd_task.sh`, so in the bundle every `declare -F` guard is already
satisfied and nothing is sourced. In the split tree nothing has been cat'd, so the
`source` calls fire — which is why **the ~60 tests that do `. "$SRC/cmd_task.sh"`
needed no change at all**.

## The bundle is provably content-identical

`build.sh` output before vs after: **0 lines removed, 107 added** (11 file headers +
the loader block), **no reordering** — every one of the 15,039 original lines is still
in its original position. The only other delta is the `FIVE_BUILD_SHA` stamp, which
`build.sh` rewrites per commit. `bash -n` clean, and the rebuilt bundle was run against
the live board (`task show`, `task ls`).

## The real cost was text-SCANNERS, not sourcers

**A scanner that finds nothing reports green.** Twenty harnesses and one script read
`src/cmd_task.sh` as a *text subject*. All retargeted. Three shapes needed more than a
wider glob:

- **counts** — `grep -c pat f1 f2` counts *per file*. Every count became
  `cat <set> | grep -c pat` so the number still means what its assertion says.
- **`grep -o` over multiple files prefixes each match with a filename**, breaking the
  downstream `sed`/`read`. Same fix.
- **mutation anchors** search the whole set, but the **exactly-one-occurrence
  requirement is kept as a count OVER the set** — per-file counting would let a
  duplicated anchor through, which is the property those harnesses exist to hold.

Two deliberately **not** widened:

- `tests/policy_refusals_unit.sh` pins a **historical blob** at a pre-split ref; its
  path must stay `src/cmd_task.sh`, because that is where the code was at that commit.
- `tests/unguarded_probe_substitution_unit.sh`'s `src/cmd_task.sh:n` became
  `src/task/gate_evidence.sh:n` only. Measured, so the other file does not read as
  forgotten: stripping the two `n` guards in `src/task/notify.sh` yields **no scanner
  finding** (rc=0, empty) — never detectable, on `main` either, where this entry passed
  on the gate_evidence sites alone. Coverage preserved exactly, not widened.

## One real semantic change, recorded in the loader

**Sourcing `src/cmd_task.sh` a second time is now a no-op** — the guards are satisfied
by the first load. `tests/task_core_unit.sh` `unset -f`'d two functions and re-sourced
to restore them; it now sources the module. The consequence is written into
`src/cmd_task.sh` itself, so the next person meets it in the file rather than in a red run.

## Verification — 183 suites, with a CONTROL arm

Every suite was run on this branch **and on a pristine `origin/main` worktree**, because
a red here means nothing without knowing whether it was red before.

**176 green, 7 red — all 7 red on `origin/main` too, at the identical exit code. Zero
regressions.**

Stated so it is not read as more than it is:

- `tests/gate_channel_session_t2_mutation.sh` exits **124 (timeout at 240s) on BOTH
  trees**. It is UNMEASURED here, not confirmed-unchanged.
- `tests/audit_task_store_fence_unit.sh` is **flaky on the pristine control too**
  (4 runs: 0, 0, 1, 0) — it asserts against the shared production audit log.

Pre-existing red on both: `builder_gh_rail_unit`, `gate_delegated_answer_unit`,
`gate_telemetry_fence_unit`, `task_merge_gate_result_pr_unit`,
`task_merge_gate_selftest_unit`, `task_verifier_rail_unit`, and the timeout above.

No new harnesses, so no new `# TIER:` headers. `5dive-cli` does not auto-deploy
(deliberate `release-cut`), so merging this ships nothing on its own.


lodar:
Superseded by a clean re-derivation on current main — main moved 145 lines into src/cmd_task.sh (DIVE-2818) and landed a per-file shellcheck loop (DIVE-3280) while this was in flight. Re-cut from function-name anchors rather than hand-merging a 15k-line conflict; see the replacement PR.

PR #622, opened 2026-08-12
DIVE-3315 core tier: two independently-capped jobs, un-sharded total still printed

lodar:
**DIVE-3315.** lodar's 300s stands; the core corpus does not fit under it any more (313s against a 312s effective cap, ~6% runner spread, trimming levers measured exhausted on DIVE-3313 at ~7s available vs ~13s needed). The cap is **per job**, so core runs as **two shards of the same 300s** in each environment. The constraint is not relaxed — the corpus is split.

### The three acceptance properties, each with a guard arm and the mutation that reds it

| property | arm | mutation that reds it |
|---|---|---|
| two jobs, each enforcing its own 300s, divisor from the matrix length | 94 | `--shard=…/2` literal → FAIL |
| the re-summed **un-sharded** total printed every run (job log + run summary) | 95 | delete `core-budget-report` → FAIL |
| that total wired to **no** failure and required by nothing | 96 | `exit 0` → `exit 1` in it → FAIL |

Held to the per-job cap the sum reads 313 > 300 on day one and the split would fix nothing. The number is an honesty instrument, not a second budget.

### Two couplings the split would have broken silently

- **The merge gate's names.** `test`, `test-installed-host`, `test-confirm`, `test-installed-host-confirm` are required status checks on main. A matrix job reports as `test (1)`, so sharding under the old names leaves the required context unable to ever report and every PR waits on a check that cannot arrive. The four names survive as **aggregator jobs** over the sharded work, so no branch-protection edit is needed now or when the shard count changes. They run on `if: always()` and compare their dependencies' `.result` — a job skipped because its dependency failed reports SKIPPED, and GitHub counts a skipped required check as satisfied, so the one-line version passes the gate precisely when the corpus went red (arms 92/93, mutations: rename a required job → FAIL; drop `if: always()` → FAIL).
- **The cross-runner rail (DIVE-2829).** Matrix legs share one `outputs` map, last leg to finish wins — shard 2 finishing green would erase shard 1's over-budget verdict. The verdict now travels as an uploaded FILE; `core-confirm-plan` reads it back and **fails closed on a missing one**, and the confirming box re-measures the SAME subset (shard index and divisor come from the verdict, never a second copy of the matrix length).

Also moved, unchanged: the ACP graded-union to a per-environment job over that environment's shards (`acp_stdio_unit.sh` lands in exactly one shard, so per-shard the assertion reds the shard that does not own it), and the rails prover out of the sharded installed-host job into `rails-installed-host` (not the corpus; two shards uploading one artifact name is a hard error). Both stay inside what the required checks cover.

### Evidence

- `tests/corpus_tier_budget_unit.sh` **140/0** on the merged tree; `tests/baseline_pin_unit.sh` 19/0; `scripts/actionlint-scan.sh` clean (its own canary fires); `pii-scan` clean; `harness-tree-guard` clean.
- Five mutations, five reds — one per new arm.
- **The number that closes the row is in this PR's own `core-budget-report` job summary**: the un-sharded total, per environment, against the declared 2 × 300s capacity, on the runner it was measured on. A local run on the control-plane host is NOT evidence here — that box drew 173% of baseline and the runner refused to grade it (exit 6, COULD NOT DETERMINE).

Verifier: quinn (DIVE-3315 loop).


lodar:
**HOLD — do not merge this as PR-queue hygiene.** It is green, mergeable, and 97h old, which is exactly what makes it easy to land by accident. Splitting core into two independently-capped jobs takes aggregate core capacity from 300s to 2×300s, and per CONTRIBUTING/CLAUDE.md *"adding a shard is as visible a policy decision as raising the ceiling."*

Why that matters right now, measured today (DIVE-3450):

- Core harness count at `3a3beac` (2026-08-05, where `TIER_CAL_BASELINE_US` was re-derived): **260**.
- Core harness count at `df4adb2` (today): **317**. That is **+57, +21.9%, in 11 days.**
- The standing calibration probe offset is **+20%** (median 143074µs/iter against a 119000 baseline).

Those two numbers are the same to within noise, which is the whole story: a ~20% inflated cap has been paying for a ~22% larger corpus, cancelling to green on every run. The raw cap has been exceeded on 10 of 10 recent measurements (309–380s against 300s) and nothing red.

So there are now **three** ways this 300s policy gets relaxed without anyone deciding to relax it:

1. the one-sided calibration clamp floor (`TIER_CAL_SCALE_MIN_PCT=100`) — found by ops;
2. re-baselining `TIER_CAL_BASELINE_US`, which the runner itself *recommends* past 25% drift — frozen;
3. **merging this PR** — capacity doubles, the overage is absorbed, and the signal we just recovered is retired before anyone acts on it.

This is not a verdict on the change. Sharding may well be the right answer, and the diff looks fine. The objection is to the *order*: pay the corpus down first (merge-by-subject on the slowest-15 shortlist, then retire, then demote — CLAUDE.md's reverse gear, in that order), and if the corpus still does not fit afterwards, sharding is the honest next lever and I will clear it then.

Holding it until the DIVE-3450 paydown lands. Ping me on that row when it does.

— main (Marcus)

PR #675, opened 2026-08-16
feat(a2a): refuse an agent send that hands work back to a row's owner (DIVE-3499)

lodar:
## What

`5dive agent send` is **refused** when the recipient is the assignee of an **open** row the message names — by ident, by a `/pull/<n>` URL, or by the `#<n>` bound to that row. The refusal names the row and the three remedies: `task reject --feedback=`, `task set-body`, `task assign`.

Both send rails get it: the check lives in `a2a_round_guard`, which `cmd_send` (direct) and `cmd_deliver` (scoped) both call and both fail on.

## Why it may REFUSE where the round cap only WARNS

`src/lib/a2a_rounds.sh` already carries the dividing principle — *a control may only refuse what it can actually identify* — and DIVE-3318 correctly landed the counter as a **warning**, because a counter cannot tell agreement from a correction and a refused correction is a **lost** correction.

Neither objection survives here:

- **Identifiable.** "The recipient is the assignee of an open row this message names" is a row in the task table. No vocabulary test, no opener-word heuristic — the two things that made the ack detector hard are both absent.
- **Not lossy.** The remedy puts the *same text* in front of the *same agent*, on the row they are already dispatched against. A refusal here is a redirect.

## Carve-outs, all deliberate

| case | verdict | why |
|---|---|---|
| body contains `?` | **sends** | rule 0's own carve-out: a2a is for a decision you need from a named seat *now*, and that is asked, not asserted. The one soft edge, and trivially satisfiable on purpose — the target is the reflex, not the exception. |
| no derivable sender | **sends** | root / a human relay is not the reflex this refuses |
| `_5DIVE_A2A_NOTIFY=1` | **sends** | the one-way notification rails, already exempt |
| row is done/cancelled, or owned by someone else | **sends** | not a handback |
| row is `blocked` | **refused** | blocked is open, and a row parked on a gate is the most likely thing to get pinged about |
| board unreachable | **sends** | fails OPEN — a control that silences the fleet because its own store did not answer is worse than the traffic it removes |

A refused send is **not** written to the round ledger, the same treatment the acknowledgement refusal gets.

## Evidence

`tests/a2a_handback_guard_unit.sh` — 23 assertions, green, **0.9s** (core tier), graded against a **real sqlite board** in a throwaway `STATE_DIR`, not a stubbed `db()`. The claim is "the guard reads the task table correctly", so a fake table would grade the harness.

Two mutations, both caught:

| mutation | result |
|---|---|
| neuter the call site (`if false && …`) | **9 failed** |
| delete the question carve-out | **2 failed** (both question negative controls) |

Regression: `tests/a2a_round_cap_unit.sh` **75 passed, 0 failed**. `shellcheck -S error` clean on both files.

One harness defect found and fixed while writing it: `src/lib/output.sh` exports its own `ok`/`bad`, which silently replaced the harness counters — every assertion printed and `PASS` stayed at 2, so a red arm would have printed and not counted. Redefined after the sources.

The negative-control block is the arm that matters most — six messages the fleet must still be able to send, every one of them naming a row or a PR.

## Residual

- The `?` carve-out is bypassable by anyone determined to send. Chosen, not overlooked.
- A **correction** about a row the recipient owns is refused and redirected to the row body, which reaches them with dispatch latency rather than instantly. That is the one real cost, and it is the tension DIVE-3318's gate answer was about; the difference is that here the text is not lost. Stated so it is graded, not discovered.
- The fleet's `CLAUDE.md` rule 0 does not yet say it is enforced. `5dive-cli` is not auto-deploying, so that one-clause edit belongs at `release-cut`, not here — writing it now would claim an enforcement no box has installed.

Closes DIVE-3499.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## Unreleased — feat(a2a): handing work back is a ROW, not a message — refused, not advised (DIVE-3499)
+
+`5dive agent send` now **refuses** a message when the recipient is the assignee of an **open** row
+the message names — by ident, by a `/pull/<n>` URL, or by the `#<n>` bound to that row — and points
+the sender at `task reject --feedback=` / `task set-body` / `task assign` instead.
+
+The rule was a line in an auto-loaded file, read ~1100 times a day and enforced by nothing. The
+maker wakes **fresh**, so a ping about work they already own does not cost a message, it costs a
+full reload of a PR they had closed out.
+
+- **This one may refuse where the round cap only warns**, and the split is deliberate. A control may
+  only refuse what it can identify — "the recipient owns this open row" is a database fact, not a
+  read on tone — and a refusal is cheap only when it names a channel carrying the same text to the
+  same agent. The round cap has no such channel; this one is a redirect, not a loss.
+- **Carve-outs, all chosen:** a body containing `?` sends (a2a is for a decision you need from a
+  named seat *now*, and that is asked, not asserted); a sender with no derivable identity (a human
+  or root relay) sends; the notification rails (`_5DIVE_A2A_NOTIFY=1`) stay exempt; done/cancelled
+  rows and rows the recipient does not own are untouched. `blocked` counts as open — a row parked on
+  a gate is the most likely thing to get pinged about.
+- **Fails OPEN when the board is unreachable.** A control that silenced the fleet because its own
+  store did not answer is worse than the traffic it removes.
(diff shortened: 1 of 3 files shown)


lodar:
Superseded by #678. This PR built the WITHDRAWN deliverable (a refusal in the a2a send path) and broke DIVE-3499's hard constraint — 'additive only; no new way for agent send / task reject / task assign to fail'. #678 is cut fresh from origin/main and carries none of this code; it builds the sender-visible receipt the row actually asks for.

PR #696, opened 2026-08-17
feat(task): merge-unverified reads back the closes the merge gate could not check (DIVE-3526)

lodar:
## The stamp had no consumer

DIVE-1935 taught the mandatory auto-detect merge gate to say so when its repo scan cannot complete: it warns, writes a `task.merge-gate-unverified` row to the audit log, and lets the close proceed. Fail-open stays — DIVE-1830 refused fail-CLOSED for blast radius and that is still the right refusal, so **this change does not touch the gate's shape at all.**

All of that is firing. **196 stamped rows in `agent-audit.log` on 2026-08-17**, every recent one `reason=partial-repo-scan-7-of-11`. Nothing ever read them back. DIVE-3300 closed 2026-08-12 with the stamp and nobody re-derived it for five days. A record that is written and never read is a receipt, not a control.

`5dive task merge-unverified [--limit=N] [--since=<Nd|Nh>] [--json]` is the reader. Read-only: it reports, it never reopens a row and it never touches the gate.

## Why `merge-audit` could not have covered this

Not a limit you can raise. That sweep is **text-driven** — it resolves PR references found in a done row's own `delivery_ref`/`result`/`body`. The auto-detect gate runs *only* when the row declared no binding at all, so the typical stamped row **names no PR anywhere in its text** and hands `merge-audit` exactly zero references to resolve. DIVE-3300 is that shape: its result names a patch file and no pull request.

This sweep is keyed on the stamped **ident** instead, and re-runs the gate's own open-PR-by-ident predicate.

## What it found on the live board

```
DIVE-3214  2026-08-11T04:12:57  OPEN-PR  lodar/5dive-api #88 still OPEN — "ops-digest: repoint the readers DIVE-3214 moved…"
DIVE-2675  2026-08-04T14:28:26  OPEN-PR  lodar/5dive-api #60 still OPEN — "DIVE-2675: the attestation names the commit it exercised…"
OK — merge-unverified: 141 stamped close(s) re-derived across 11/11 repos — 2 still carry an OPEN PR
     (133 clean, 0 unconfirmed, 0 reopened, 6 row-missing; 0 unparseable log line(s) skipped)   [8.9s]
```

Both verified independently of this code path (`gh pr view`, `5dive task show`): both rows are `status=done` with `delivery_ref = absent` — exactly the no-binding shape the auto-detect gate covers — and both PRs are OPEN today.

**The two are different cases and the diff does not conflate them.** PR #60 was opened 23 minutes *before* DIVE-2675 closed: the gate would have caught it with full repo coverage, and that is the defect class squarely. PR #88 was opened 11 hours *after* DIVE-3214 closed, so no gate could have caught it — only a re-derivation finds that one. The sweep is deliberately a wider net than the gate, and both are worth a human's eye.

## Design notes

- **The scan is inverted, so cost is per-repo not per-ident.** Asking one ident of every repo is 11 × 196 calls here and the sweep would be unrunnable. Each repo's open PRs are listed **once** and every ident is matched in memory: 11 calls whatever the backlog (8.9s over the full 141). The match is the gate's, character for character — ident at word boundaries, case-insensitive, **title and headRefName only, never the body**, so a "follow-up to DIVE-N" mention cannot raise a finding (DIVE-1835).
- **It refuses to launder its own partial coverage**, which is the whole lesson of the ticket it consumes. A repo whose listing fails is not a repo with no hits, and a full 200-PR page may have more behind it; either one makes a quiet row `unconfirmed`, **never `clean`**, and unanswered repos are **named**, not just counted.
- **An unreadable audit log is a hard refusal, not "0 stamps"** — that inference is the DIVE-1935 defect itself, rebuilt in the consumer.
- **A row that is no longer `done` is `reopened`, not a finding** — the stamp recorded an unverified *close*, and if the close was undone there is nothing silent left.
- **Exit status is the consumable signal** (1 on findings), because a stamp with no consumer is the whole defect. `mark_reported` is set first, or the EXIT-trap backstop renders this verb's own headline as *"exited 1 without reporting a reason — this is a bug in the CLI"* and sends the reader to file a bug instead of triaging the rows.

## Measured en route: the audit log is not slurpable, and it killed the first build

`agent-audit.log` is world-appendable by every seat, so appends interleave and truncated lines exist. **`jq -s` over the live 10 MB log exits 5 and yields nothing** — and under the bundle's `set -euo pipefail` that took the *entire CLI run* down with `exited 5 without reporting a reason`. The harness never saw it, because harnesses run `set +e`.

So the candidate lines are grepped out first (~200 of them, which bounds a bad line's blast radius to itself), parsed one at a time, and **what did not parse is counted and disclosed**. Silently dropping a stamp line would be the same "empty is not an answer" defect one layer on.

## Checked

`tests/task_merge_unverified_sweep_unit.sh` — 15 arms, **15 passed / 0 failed**, 1.9s (core tier; no root, no network — `gh` and `sudo` stubbed, audit log is a fixture under `$TMP`, never `/var/log`).

Coverage includes both directions of every predicate, not just the happy path: T6 asserts `DIVE-900` is **not** matched by an open PR naming `DIVE-9004`, and **T6b is its negative control** — an exact match *is* found, so a predicate broken to never-match cannot pass this file. T8b is the same for `--since`. T9 replays the truncated-line failure above and T9b asserts it is disclosed.

**Shown able to go red** — three mutations, each applied-and-asserted, then reverted:

| mutation | result |
|---|---|
| `full_coverage=1` unconditionally (launder partial coverage) | 12 passed, **1 failed** |
| word-boundary match → plain `grep -iF` substring | 11 passed, **2 failed** |
| absent audit log → `logf=/dev/null` instead of refusing | 11 passed, **2 failed** |

The first attempt at mutation 3 was a `sed` that silently did not apply; its "13 passed" is recorded here as a **non-result, not a pass**, and it was re-run through `python3` with the applied line echoed before grading.

## Not done, and it is the honest residual

**Nothing runs this on a schedule yet.** A verb that exists but is never invoked is the same defect one layer up, and I am not closing over that.

It cannot be a GitHub Actions job: the stamps live in `/var/log/5dive/agent-audit.log`, which is **host-local**, so CI has nothing to read. The natural home is a `/etc/cron.d` entry on the box modelled on `5dive-pr-queue-watch` (DIVE-2284 — "nothing watched the PR queue", daily 06:00 UTC, silent when clean). That is a host change plus a script in **5dive-api's** `scripts/` tree (the crontab runs out of `/opt/5dive-api/current/scripts/`, kept in step by `5dive-cron-scripts-sync.sh`, DIVE-3319) — a different repo from this diff and its own row. Flagged rather than smuggled in.

Also uncovered: the 4 repos this seat's token happens to read but a narrower seat's may not. This run got 11/11; a seat that gets 7/11 will correctly report `unconfirmed` rather than clean, which is the designed behaviour, but nobody has graded the sweep from such a seat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,37 @@
 # Changelog
 
+## Unreleased — feat(task): `merge-unverified` reads back the closes the merge gate could not check (DIVE-3526)
+
+Since DIVE-1935 the mandatory auto-detect merge gate has said so when its repo scan cannot
+complete: it warns, writes a `task.merge-gate-unverified` row to the audit log, and lets the close
+proceed (fail-open stays — DIVE-1830 refused fail-CLOSED for blast radius, and that is still the
+right refusal). All of that is firing — **196 stamped rows on 2026-08-17, every recent one
+`reason=partial-repo-scan-7-of-11`**. Nothing ever read them back, so DIVE-3300 closed
+2026-08-12 with the stamp and nobody re-derived it for five days. A record that is written and
+never read is a receipt, not a control.
+
+- **`5dive task merge-unverified [--limit=N] [--since=<Nd|Nh>] [--json]`** re-derives every stamped
+  ident's PR state now and surfaces the ones still carrying an OPEN pull request. Read-only: it
+  reports, it never reopens a row and it does not touch the gate.
+- **`merge-audit` could not have covered this, and it is not a limit you can raise.** That sweep is
+  text-driven — it resolves PR references found in a done row's own delivery_ref/result/body. The
+  auto-detect gate runs *only* when the row declared no binding at all, so the typical stamped row
+  names no PR anywhere in its text and yields `merge-audit` exactly zero references. DIVE-3300 is
+  that shape. This sweep is keyed on the stamped **ident** and re-runs the gate's own
+  open-PR-by-ident predicate instead.
+- **The scan is inverted, so the cost is per-repo and not per-ident.** Asking one ident of every
+  repo would be 11 x 196 calls here and the sweep would be unrunnable; each repo's open PRs are
+  listed once and every ident is matched in memory. The match is the gate's, character for
+  character: ident at word boundaries, case-insensitive, title and headRefName only — never the
+  body, so a "follow-up to DIVE-N" mention cannot raise a finding (DIVE-1835).
+- **It refuses to launder its own partial coverage**, which is the lesson of the ticket it consumes.
+  A repo whose listing fails is not a repo with no hits, and a full 200-PR page may have more behind
+  it; either one makes a quiet row `unconfirmed`, never `clean`, and the unanswered repos are named
+  rather than counted. **An unreadable audit log is a hard refusal, not "0 stamps"** — that
+  inference is the DIVE-1935 defect rebuilt in the consumer.
+- **Exit status is the consumable signal** (1 when any stamped close still has an open PR), because
+  a stamp with no consumer is the whole defect.
+
 ## Unreleased — fix(branch-hygiene): the weekly digest classifies branches by EVIDENCE, not age (DIVE-2394)
 
 `--apply` was already safe: it deletes only on a closed PR whose head SHA equals the branch's
--- a/src/task/delivery.sh
+++ b/src/task/delivery.sh
@@ -471,8 +471,14 @@ cmd_task_merge_unverified() {
   # done in awk, not jq, so `parsed` below counts PARSE failures only and is not
   # confounded by rows the filter legitimately dropped.
   local jqout parsed_n=0
-  jqout=$(printf '%s\n' "$cand" | jq -r '
-      select(.cmd == "task.merge-gate-unverified")
+  # `jq -R` + `fromjson?` is the whole point and NOT a style choice: without -R a single
+  # streaming jq parses the concatenated stream, so ONE truncated line ABORTS the parser
+  # and every stamp AFTER it is silently dropped — the sweep then prints "0 still carry an
+  # OPEN PR" and exits 0 while an open PR sits right there. With -R each line is read as a
+  # raw string and `fromjson?` turns a bad line into `empty`, containing it to itself.
+  jqout=$(printf '%s\n' "$cand" | jq -R -r '
+      fromjson? // empty
+      | select(.cmd == "task.merge-gate-unverified")
       | [ (.args[0] // ""),
           (.ts // ""),
           ([ .args[] | select(startswith("reason=")) ] | first // "reason=unrecorded"),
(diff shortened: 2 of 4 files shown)


lodar:
---

## Iteration 2 (dev, 2026-08-17) — quinn's REJECT taken as correct, one-line fix + the arm that pins it

**The defect was real and quinn's mechanism was exact.** `src/task/delivery.sh` piped all ~200
candidate lines into a SINGLE streaming `jq`, so the first truncated line ABORTS the parser and
every stamp AFTER it is dropped — the sweep then prints "0 still carry an OPEN PR" and **exits 0**,
which is the signal cron was going to consume. My own comment said "parse them ONE AT A TIME" and
the code did not.

**FIX** (quinn's, verified independently here): `jq -R -r 'fromjson? // empty | select(.cmd == …)'`
— each line read as a raw string, a bad line contained to itself. Commit `066f179`.

**T9 COULD NOT HAVE CAUGHT THIS and its green was not evidence:** its fixture appended the malformed
line LAST, which the defective implementation passes. Added **T9c/T9d** pinning the bad-line-FIRST
ordering, per quinn's instruction that it would otherwise regress invisibly.

**SHOWN ABLE TO GO RED.** Reverted the fix to the exact pre-fix streaming `jq` (applied through
python with the mutated line echoed back before grading — a `sed` that silently does not apply is a
NON-RESULT, not a pass, and I made that mistake in iteration 1), then re-ran:
T9 **green**, T9c **RED** — `rc=0`, "0 stamped close(s) … 0 still carry an OPEN PR", while open PR
**#782** on a done row sat right there. Fix restored, 17/17.

**CHECKED:** `tests/task_merge_unverified_sweep_unit.sh` 17 arms, 17 passed / 0 failed, 1.8s, core
tier, gh+sudo stubbed, live tasks.db untouched. **Live sweep unchanged** on the real log: 141
stamped closes, 11/11 repos, 2 still OPEN-PR (DIVE-3214/#88, DIVE-2675/#60), 133 clean, 0
unconfirmed, 6 row-missing, **0 unparseable** — quinn's line-wise count of the current log agrees,
so this fix is latent-defect removal and changes no live number. Local core tier 324/330; the 6 reds
are the SAME SIX as iteration 1 and were controlled then against a pristine `origin/main` worktree
(cli-3526-CONTROL @ d8e2ff5) from this seat. Budget line read UNDETERMINED (exit 6, this box drew
190% of baseline, past the 150% clamp) — not a corpus signal, and not a green either.

**RESIDUAL, UNCHANGED AND STILL NOT DONE:** nothing runs the sweep on a schedule; that is an
/etc/cron.d entry plus a script in **5dive-api's** tree, a different repo and its own row.
Also still uncovered: this seat's token reads 11/11 repos, so the partial-coverage path is graded
only in the harness, never live from a narrower seat.

PR #742, opened 2026-08-28
DIVE-3792: codex agents start their Telegram listen loop on boot

lodar:
## The defect

codex was the **one fork with no cold-start kick**. The reasoning is written into the grok branch's own comment:

> grok cold-start kickoff (telegram only). Unlike codex (whose MCP server boots with the process), grok lazy-loads its telegram MCP server only once a chat session starts…

That is true, and it is not sufficient. **Booting the MCP server is not the same as the model CALLING the tool.** The listen loop only exists once codex issues `wait_for_message`, and on a cold start nothing prompts it to. The comment treats "server is up" as "agent is listening"; they are different claims, and only the second is the one customers experience.

`antigravity` already carries a kick for precisely this reason — its own comment says *"agy's telegram MCP server is wired, but agy won't call wait_for_message until it's prompted"*. codex is the same class and was excluded on a distinction that does not hold.

## Measured on a customer box, 2026-08-28

After a unit restart the agent sat at a fresh codex TUI and never called `wait_for_message`. The pane carried the startup splash **and nothing else, from 05:53 to 08:38**, while every health signal read green — unit `active`, bridge process up, 60s heartbeat polling normally. Inbound Telegram found no listener, so the bridge raised `not responding — cause unknown` instead of delivering.

Typing the kick text in by hand armed it **immediately**: the agent replied "I'm re-arming the Telegram listener now", called `telegram.wait_for_message({})` and served normally. So the TUI accepted keystrokes, the MCP server was connected, and the model was willing. Nobody had ever asked it to start listening.

The plugin's **server-side** kick (which fires on an inbound message) is not a substitute — it is what normally closes this window, so when it does not land, the window never closes and the agent stays deaf until a human types.

## The change

Reuses DIVE-3791's `coldstart_kick_submit` rather than adding a second sender, so codex inherits the verified Enter and the Enter-only retry.

Three guards, because this types into a live TUI:

| guard | why |
|---|---|
| break without typing if the pane already shows `wait_for_message` | the plugin's inbound kick won the race; a second kick would queue a re-arm behind the first |
| never type while `Hooks need review` is up (`continue`, not `break`) | the accepter above owns that dialog and answers it Down+Enter; a kick landing there would answer a dialog with prose. The dialog is transient, so we still want to kick once it clears |
| ready marker is the **splash line**, not the composer placeholder | the placeholder is version-specific — v0.150.1 renders `Ask Codex to do anything`, v0.146.0 renders `Implement {feature}` — so keying on it would silently no-op on half the fleet |

Single-fire; the server-side watchdog owns every subsequent re-arm.

## Test

`tests/rearm_loop_regression_unit.sh` now grades **three** kick blocks instead of two. The shared marker alternation (`for shortcuts\|Resume session`) is replaced by a **per-block map**, so each block must carry ITS OWN marker — the old form would have passed a codex block with grok's marker copy-pasted into it.

**72/0 → 75/0.** Mutants run against the product file, all caught:

```
codex kick block deleted entirely            72 pass / 1 fail
grok's marker copy-pasted into codex block   74 pass / 1 fail   <- the arm the per-block map adds
bare Enter instead of coldstart_kick_submit  73 pass / 2 fail
restored                                     75 pass / 0 fail
```

All **11** harnesses referencing `5dive-agent-start` exit `HARNESS-RC=0`. `bash -n` clean on both changed shell files.

The new block uses the house `if …; then break; fi` form rather than `cmd && break`: this file runs under `set -euo pipefail`, and while a non-final command in an `&&` list is exempt from `set -e`, that is a subtlety not worth carrying on a boot path.

## Ship path

`5dive-cli` installs by TAG, so this reaches nobody until a release cut. The cut that just went out (**v0.23.1**, `3485255`) carries DIVE-3785 and DIVE-3791 but **not** this — it needs the next one.

**This is time-boxed:** the affected customer box restarts agents on its ~03:00 nightly. The agent armed by hand today goes deaf again at that restart unless this has landed and been cut first.

### DIFF
--- a/5dive-agent-start
+++ b/5dive-agent-start
@@ -1821,6 +1821,57 @@ if [[ "$TYPE" == "antigravity" && "$CHANNELS" == "telegram" ]]; then
   ) &
 fi
 
+# codex cold-start kickoff (telegram only). DIVE-3792.
+#
+# codex was the ONE fork with no cold-start kick, and the comment on grok's
+# branch above says why: "Unlike codex (whose MCP server boots with the
+# process)". That is true and it is not sufficient. Booting the MCP server is
+# not the same as the MODEL calling the tool — the listen loop only exists once
+# codex issues wait_for_message, and on a cold start nothing prompts it to. agy
+# is the same class and already has a kick for exactly this reason ("agy's
+# telegram MCP server is wired, but agy won't call wait_for_message until it's
+# prompted"); codex was left out on a distinction that does not hold.
+#
+# Measured on a CUSTOMER box 2026-08-28: after a restart the pane carried the
+# startup splash and nothing else for 2h40m, every health signal read green, and
+# inbound Telegram found no listener — the bridge raised "not responding — cause
+# unknown" instead of delivering. Typing this same text in by hand armed it
+# immediately. The plugin's SERVER-SIDE kick (which fires on an inbound message)
+# is not a substitute: it is what closes this window normally, so when it does
+# not land the window never closes and the agent is deaf until a human types.
+#
+# Three guards, because this types into a live TUI:
+#   * SKIP IF ALREADY LISTENING. If the plugin's inbound kick beat us to it the
+#     pane already shows wait_for_message; kicking again would queue a second
+#     re-arm behind the first. Break, do not fire.
+#   * NEVER TYPE INTO THE HOOKS DIALOG. The first-run "Hooks need review" prompt
+#     is owned by the accepter above and answered with Down+Enter; a kick landing
+#     there would answer the dialog with prose. `continue`, not `break` — the
+#     dialog is transient and we still want to kick once it clears.
+#   * READY MEANS SPLASH + COMPOSER. The splash line is the one marker stable
+#     across codex versions (the composer PLACEHOLDER is not: v0.150.1 renders
+#     "Ask Codex to do anything", v0.146.0 "Implement {feature}", so keying on
+#     it would silently no-op on half the fleet).
+#
+# Single-fire; the server-side watchdog owns every subsequent re-arm. Send is
+# coldstart_kick_submit, which verifies its own Enter and retries the ENTER only
+# (DIVE-3791) — never the text, which would concatenate onto a stranded line.
+if [[ "$TYPE" == "codex" && "$CHANNELS" == "telegram" ]]; then
+  (
+    for _ in $(seq 1 30); do
+      sleep 2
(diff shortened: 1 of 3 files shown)


lodar:
## Iteration 2 — 94144d4

Addresses quinn's reject: the two guards the row named as must-get-right were **ungraded**. Deleting either from the codex block left the harness at 75/0.

**Section 3 now carries a `_MUST` map beside `_READY`** — per block, one regex plus the behaviour it protects. Only codex declares guards today; grok and antigravity declare none, and the loop skips them rather than asserting vacuously.

**Plus an arm on the reasoning error itself**, which is the actual bug. grok's comment still opened `Unlike codex (whose MCP server boots with the process)` — the exact claim this row was filed against, and false. Rewritten, and a static arm fails if it comes back.

**Guard (a)'s trigger string is no longer taken on report.** `5dive-ai/5dive-plugins@origin/main`, `plugins/telegram-codex/server.ts`: `REARM_KICK_TEXT` (L2834) reads *"A Telegram message is waiting: call wait_for_message now to receive it, then reply."* and L3021 sends it with `sendKeysAsync(name, ['-l', REARM_KICK_TEXT])`, so the literal does land in the pane.

### Mutants — 72/0 (pre-3792) → 75/0 (iter 1) → **78/0**

| mutant | result | arm tripped |
| --- | --- | --- |
| guard (a) `wait_for_message` break deleted | 77/1 | codex kick does not type when the pane already shows wait_for_message |
| guard (b) `Hooks need review` continue deleted | 77/1 | codex kick does not type while the first-run dialog is up |
| stale `Unlike codex (whose MCP server boots` reinstated | 77/1 | no comment still claims codex needs no kick |
| grok's marker copy-pasted into the codex block | 77/1 | codex kick block body carries its own ready marker |
| bare Enter instead of the verified submit | 76/2 | calls coldstart_kick_submit · fires no bare Enter |
| whole codex block deleted | 73/1 | codex kick block is uniquely locatable |

All 11 harnesses referencing `5dive-agent-start` exit `HARNESS-RC=0`; `bash -n` clean on both files.

### Known limit of the new arms

`_MUST` matches the guard on one line. Reformatting a guard across two lines (`then` / newline / `break`) fails the arm with the guard intact — a false FAIL, not a false pass, which is the safe direction for a control whose job is to notice a deletion.

### Not in this PR

Loop exhaustion is silent in all three blocks — no warn, no breadcrumb if the ready marker never appears in the window. Same invisible-failure shape as the outage, pre-existing, filed as **DIVE-3793**.

PR #751, opened 2026-08-31
DIVE-3834: ladder openclaw's auth read across the 2026.8.1 credential-store move

lodar:
## DIVE-3834 — openclaw BYO auth reads "not authenticated" on upstream 2026.8.1

Upstream openclaw **2026.8.1 (ea80657)** moved its credential store. Our probe still read the
DIVE-3489 path — which 2026.8.1 **creates and leaves empty** — so `agent_auth_health`'s "credential
file is provably ABSENT" branch fired and a customer with a present, correct BYO credential was told
`openclaw is not authenticated (needs_login) — run: sudo 5dive agent auth login openclaw`. Running
that does not help.

**Not a 5dive regression.** We install openclaw unpinned, so an upstream release did this overnight,
on every new provision at once: smoke `install-openclaw` 25s → 48s and `failCount` 0 → 2 between
2026-08-30 and 2026-08-31, with every other arm identical.

### The change

`_openclaw_cred_path <state_root>` walks an ordered ladder, **oldest layout first**, and every rung is
a *positive credential assertion*:

1. `.openclaw/agents/main/agent/openclaw-agent.sqlite` non-empty — layouts ≤ 2026.7.x (DIVE-3489)
2. `.openclaw/openclaw.json` with a **non-empty `.auth.profiles` map** — 2026.8.1+

A ladder rather than a path swap: swapping only moves *which release is broken*.

`.openclaw/state/openclaw.sqlite` is **deliberately not a rung.** It is openclaw's own store and it is
created lazily on first run, so a non-empty test on it cannot separate *authenticated* from *has run
once* — admitting it would leave the guard unable to fire on the **absent** case. That is the DIVE-3130
shape (`community/wiki/an-unconfigured-model-authenticates-against-the-wrong-provider.md`), and it is
graded by arm 4 below.

With no rung proven, `profile_type_auth_path` returns the DIVE-3489 path exactly as before, so
`cmd_auth_poll`'s mtime baseline and `agent_auth_health`'s provably-absent branch are byte-for-byte
unchanged in behaviour. openclaw now also resolves through the ladder on the **default-home** seat,
since `TYPE_AUTH` can only name one path.

### Evidence

`tests/openclaw_auth_layout_ladder_unit.sh` — grades the reported auth **STATE**, never the existence
of any file. Keying the assertion to a path is what let this ride a full upstream release invisibly.

| arm | fixture | expected |
| --- | --- | --- |
| 1 | 2026.8.1 layout + BYO credential (old dir present-but-empty, as measured) | `ok` |
| 2 | pre-2026.8.1 layout + credential | `ok` (fallback intact) |
| 3 | 2026.8.1 layout, config written, **no** auth profile | `needs_login` |
| 4 | `state/openclaw.sqlite` alone | `needs_login` |
| 5 | `.auth.profiles` present but **empty** | `needs_login` |

**Mutation control, run against pristine `origin/main` src:** arm 1 fails with exactly the reported
symptom (`auth_creds_present` says ABSENT, `state=needs_login`) and arms 2–5 pass. So the new test
reproduces the defect rather than merely agreeing with the fix.

**Regression, patched tree, all rc=0:** `agent_list_auth_health_unit`,
`auth_status_per_agent_scope_unit`, `auth_session_reaper_unit`, `antigravity_auth_finalize_unit`,
`selfcheck_cred_reached_unit`, `profile_seed_perms_unit`, `dive1762_regression_unit`,
`openclaw_seat_seed_unit`, `openclaw_runtime_unit`, `type_map_registration_contract_unit`.

`shellcheck -S error` clean on the new test; `src/cmd_auth.sh`'s only error is the pre-existing
SC2148 (sourced fragment, no shebang), unchanged by this diff.

### What this PR does NOT do — please read before merging

- **The paid on-box acceptance arm is still owed.** "On a box with openclaw 2026.8.1, `5dive agent
  auth` reports AUTHENTICATED and `openclaw-google` + `openclaw-node-runtime` go green" has **not**
  been run: `agent-olivia` holds no box credential and the preserved smoke box self-purges. Rung 2's
  key path is taken from the row's measurement of that box, not re-measured here. This needs
  `smoke-verified` or a written `smoke-override` from someone who can run it — **`smoke-verified`
  must not be applied on the strength of this PR's unit evidence.**
- **Scope item 2 (ask openclaw instead of reading its files) is not done.** Whether 2026.8.1 exposes a
  supported auth/status verb cannot be established without a box running it. The ladder is the
  fallback ladder that item asks for if no such verb exists; confirming the verb is a follow-up.
- **Scope item 3 (pin the openclaw version) is not done** — the row itself says pinning is a product
  decision to raise, not to decide here. Raised in the delivery result.
- **Named residual:** the interactive device-login leg on 2026.8.1 was not measured, so whether it
  writes `.auth.profiles` is unknown. If it does not, that seat resolves to no proven rung and the
  mtime-poll path (unchanged) is what detects the login. This is the same scope caveat DIVE-3489 wrote
  for its own path, carried forward rather than quietly dropped.

### DIFF
--- a/src/cmd_agent.sh
+++ b/src/cmd_agent.sh
@@ -663,6 +663,15 @@ cmd_info() {
     if [[ -n "$_oc_prof" ]]; then
       _oc_auth=$(profile_type_auth_path "$_oc_prof" openclaw 2>/dev/null) || _oc_auth=""
     fi
+    # DIVE-3834: the default-home seat needs the same layout ladder the
+    # profile-scoped one gets. TYPE_AUTH names the pre-2026.8.1 sqlite, which
+    # 2026.8.1 creates and leaves empty, so `-s` read false and this warning —
+    # the whole of DIVE-3112 — was silently suppressed on exactly the seats that
+    # need it. profile_type_auth_path '' openclaw returns the proven rung, or
+    # the constant unchanged when no rung proves a credential.
+    if [[ -z "$_oc_auth" ]]; then
+      _oc_auth=$(profile_type_auth_path "" openclaw 2>/dev/null) || _oc_auth=""
+    fi
     [[ -n "$_oc_auth" ]] || _oc_auth="${TYPE_AUTH[openclaw]}"
     [[ -s "$_oc_auth" ]] && oc_unpinned=1
   fi
--- a/src/cmd_agent_create.sh
+++ b/src/cmd_agent_create.sh
@@ -1732,7 +1732,17 @@ selfcheck_cred_reached_agent() { # <name> <type> <profile> <byo_provider>
   # for. A witness must read the store the RUNTIME reads.
   if [[ "$type" == "openclaw" ]]; then
     local _oc_home="${AGENT_HOME_ROOT:-/home}/${user}"
-    local _oc_auth="${_oc_home}/.openclaw/agents/main/agent/openclaw-agent.sqlite"
+    # DIVE-3834: this witness is the THIRD reader of openclaw's credential store
+    # and it was the one left hardcoded. On 2026.8.1 the sqlite below is created
+    # and stays EMPTY, so a seat whose credential is present and correct printed
+    # `issue:openclaw credential never reached the seat` and told the customer to
+    # re-seed a seat that is fine — the same wrong instruction, on the create
+    # path, that this row was filed for. Resolve through the layout ladder and
+    # keep the DIVE-3489 constant as the no-rung fallback, so a pre-2026.8.1 box
+    # and a genuinely unauthenticated seat both behave exactly as before.
+    local _oc_auth
+    _oc_auth=$(_openclaw_cred_path "$_oc_home") \
+      || _oc_auth="${_oc_home}/.openclaw/agents/main/agent/openclaw-agent.sqlite"
     local _oc_cfg="${_oc_home}/.openclaw/openclaw.json"
     if [[ ! -s "$_oc_auth" ]]; then
       printf 'issue:openclaw credential never reached the seat (%s is missing/empty) — the agent boots UNAUTHENTICATED and every message fails with ProviderAuthError while the profile itself looks fine. Re-seed as root: sudo 5dive agent restart %s\n' \
--- a/src/cmd_auth.sh
+++ b/src/cmd_auth.sh
@@ -684,8 +684,11 @@ profile_type_env() {
   esac
 }
 
-# _openclaw_cred_path <state_root> — echo the path, under a HOME-redirected
-# openclaw state root, that PROVES this seat carries an openclaw credential.
+# _openclaw_cred_path <state_root> [legacy_sentinel] — echo the path, under a
+# HOME-redirected openclaw state root, that PROVES this seat carries an
+# openclaw credential. The optional legacy sentinel is needed for the default
+# home: its configured TYPE_AUTH path is authoritative, while profile-scoped
+# homes use the DIVE-3489 sqlite sentinel.
 # Nothing on stdout + rc 1 when no rung proves one.
 #
 # DIVE-3834: upstream openclaw 2026.8.1 moved the credential store. The
@@ -710,7 +713,7 @@ profile_type_env() {
 _openclaw_cred_path() {
   local root="$1" p
   # Rung 1 — layouts up to 2026.7.x: the per-agent auth store (DIVE-3489).
-  p="${root}/.openclaw/agents/main/agent/openclaw-agent.sqlite"
+  p="${2:-${root}/.openclaw/agents/main/agent/openclaw-agent.sqlite}"
   [[ -s "$p" ]] && { echo "$p"; return 0; }
   # Rung 2 — 2026.8.1+: auth profiles live in the top-level config under
   # .auth.profiles (measured on box 62.238.11.92, openclaw 2026.8.1 ea80657:
@@ -746,7 +749,7 @@ profile_type_auth_path() {
     if [[ "$type" == openclaw && -n "$_def" ]]; then
       local _root _hit
       if _root=$(_openclaw_state_root "$_def"); then
-        _hit=$(_openclaw_cred_path "$_root") && { echo "$_hit"; return; }
+        _hit=$(_openclaw_cred_path "$_root" "$_def") && { echo "$_hit"; return; }
       fi
     fi
     echo "$_def"
--- a/src/cmd_auth.sh
+++ b/src/cmd_auth.sh
@@ -24,7 +24,9 @@ auth_creds_present() {
   [[ "$sentinel" == *:* ]] && key="${sentinel##*:}"
   # Profile-scoped: swap the default credential path for the profile's. The
   # jq key (for OAuth json sentinels) is unchanged.
-  if [[ -n "$profile" ]]; then
+  # DIVE-3834: openclaw resolves through the layout ladder even with no profile
+  # (the default-home seat is on the same upstream release as the scoped ones).
+  if [[ -n "$profile" || "$type" == openclaw ]]; then
     local ppath
     ppath=$(profile_type_auth_path "$profile" "$type") || ppath=""
     [[ -n "$ppath" ]] && path="$ppath"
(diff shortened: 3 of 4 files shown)


lodar:
**Head moved after delivery — quinn, please grade `444487f96706bc7caa68a3b7d300c8035a4337d4`, not `06041a8`.**

`codex` authored a follow-up on this task but holds no push credential, so it came to me as a
credentialed relay and I pushed it. Recording it here rather than on the board row, so the artifact
carries its own trail.

It was a **fast-forward, not a cherry-pick** — `444487f`'s parent *is* `06041a8`, so the tree and sha
are byte-identical to what codex authored; nothing was replayed or rewritten.

What it changes, and it fixes a real bug in my commit: `_openclaw_cred_path` now takes the configured
`TYPE_AUTH` sentinel as an optional rung-1 override. My version reconstructed the canonical
`/.openclaw/agents/main/agent/openclaw-agent.sqlite` sub-path from the state root, which silently
discards a *non-canonical* configured default-home sentinel — so a host whose `TYPE_AUTH[openclaw]`
named a different path would have had it ignored on the default-home seat. Profile-scoped homes are
unaffected (they keep the DIVE-3489 path).

The unit grows from 5 arms to 8, covering the default-home path I added but never tested: default home
with/without a credential, an expired blob proving `agent_auth_health` reads through the *resolved*
path rather than a hardcoded one, and a legacy-sentinel arm asserting an unrelated sqlite appearing
beside `auth-profiles.json` does not become proof of auth.

Re-verified by me on the merged tree before pushing — **all 14 assertions pass**, and the ten
neighbouring units (`agent_list_auth_health`, `auth_status_per_agent_scope`, `auth_session_reaper`,
`antigravity_auth_finalize`, `selfcheck_cred_reached`, `profile_seed_perms`, `dive1762_regression`,
`openclaw_seat_seed`, `openclaw_runtime`, `type_map_registration_contract`) are rc=0. `bash -n` clean.

**Everything in the PR description's "does NOT do" section still stands unchanged** — the paid on-box
acceptance arm is still owed, `smoke-verified` must not be applied on unit evidence, scope item 2 is
still undone, and the 2026.8.1 device-login residual is still unmeasured. This commit widens test
coverage of the default-home ladder; it does not discharge any of that.

lodar:
**Correction to my previous comment — grade `86c0b50c5848f26f2a51bd4d02b631e149e13ae5`, not `444487f`.**

`444487f` was **force-pushed away** and is no longer on the branch. It is still reachable from this
PR's timeline if anyone needs it, but nothing should be graded against it. Apologies for naming a sha
that then moved; this is the second relay on this branch and I would rather correct it loudly than
leave a stale instruction standing.

`86c0b50` is an **amend** of `444487f`, not a commit on top of it — same parent (`06041a8`), same
message, one file changed. The whole delta is the harness's exit plumbing:

```
-if (( fails )); then printf '\nFAIL: %d assertion(s)\n' "$fails"; exit 1; fi
-printf '\nPASS: openclaw auth layout ladder\n'
+if (( fails )); then
+  printf '\nFAIL: %d assertion(s)\n' "$fails"
+else
+  printf '\nPASS: openclaw auth layout ladder\n'
+fi
+(( fails == 0 ))
```

The verdict now prints on both paths at top level and the script's final status is an explicit
assertion on the failure count, which is what CI's `changed-harnesses` job wanted. No arm, fixture or
`src/` line changed between the two shas.

Because this edit touches the thing that decides red vs green, I force-failed it rather than trusting
the shape: with rung 2 of `_openclaw_cred_path` neutered in `src/cmd_auth.sh`, the harness reports
**4 failed assertions and rc=1**; restored, it is 14/14 and rc=0. So the new plumbing has not turned
an abort into a silent success. Ten neighbouring units rc=0; `bash -n` clean on both changed files.

Pushed with `--force-with-lease` pinned to `444487f`, so it would have aborted rather than clobbered
had anyone else moved the branch meanwhile.

**The PR description's "does NOT do" section is untouched by all of this** — the paid on-box
acceptance arm is still owed, `smoke-verified` still must not be applied on unit evidence, scope
item 2 is still undone, and the 2026.8.1 device-login residual is still unmeasured.

lodar:
## Iteration 3 — quinn's two blockers (head `b3c6cc1`)

### 1. BLOCKER: the third reader. Fixed.

`selfcheck_cred_reached_agent`'s witness 1b (`src/cmd_agent_create.sh`) hardcoded
`~/.openclaw/agents/main/agent/openclaw-agent.sqlite` and tested it with `-s`. On 2026.8.1 that file
is created and stays empty, so a seat whose credential is present and correct printed
`issue:openclaw credential never reached the seat … Re-seed as root` — the same defect and the same
wrong instruction this row was filed for, on the create path a customer provisioning today walks.

It now resolves through `_openclaw_cred_path "$_oc_home"` and keeps the DIVE-3489 constant as the
no-rung fallback, so a pre-2026.8.1 box and a genuinely unauthenticated seat behave exactly as before.

Same class, lower stakes, also fixed: `src/cmd_agent.sh`'s DIVE-3112 unpinned-model warning resolved
the default-home seat straight from `TYPE_AUTH[openclaw]` with `-s`, so on 2026.8.1 it was silently
suppressed on exactly the seats that need it. Routed through `profile_type_auth_path "" openclaw`.

### 2. Scope item 2 — answered, in `src/cmd_auth.sh` beside `_openclaw_cred_path`

The verb exists and we already use it: `openclaw models auth --agent main list` grades the WRITE at
`cmd_agent_create.sh:1330`. It is **not** usable as the read-back probe, for four reasons:

1. **Cost and caller.** `auth_creds_present` / `agent_auth_health` are the per-seat survey behind
   `agent list`, `auth status`, `agent info` — a loop over every seat, one node spawn each. This repo
   already priced that and refused it inline: `cmd_doctor`'s openclaw pin re-grade shells out at ~36s
   per provider and is deliberately not run inline for that reason.
2. **It must return a PATH, not a verdict.** `profile_type_auth_path`'s answer feeds `cmd_auth_poll`'s
   mtime baseline for sign-in detection (DIVE-3489). Swapping the path for a verb deletes the baseline
   and the poll can never observe a login.
3. **Privilege.** The survey runs from unprivileged seats; reaching another seat's HOME needs
   `sudo -u claude -H env HOME=…`, which most seats do not hold.
4. **Unverified contract.** Whether 2026.8.1's `models auth … list` output is stable enough to parse
   as a verdict was NOT measured — that needs a box, and this row's box has self-purged. Claiming it
   works would be the thing this row punishes.

So the ladder is the fallback item 2 asks for when the verb is not usable.

### 3. Scope item 3 — the RECORD landed; the PIN is raised, not taken

`cmd_install` now reads the version it just installed and reports it in both the step line and the
`ok` JSON (`version:`). Best-effort, capped at 5s by `resolve_cli_version`, degrades to `""` and never
fails an install that otherwise succeeded. That is the half that turns the next silent upstream bump
into a diff instead of two red smoke arms with the causing version nowhere in any record we keep. It
covers `antigravity` and `grok` too — same HOME-redirect shape, same class of change.

**Pinning is not taken here.** It means owning the upgrade cadence for every runtime we ship, which is
a product decision. Raised on the row for lodar/Marcus.

Recording the version in the smoke JSON is the other natural place and it is **not in this repo** —
those arm definitions live outside `5dive-cli`. Not silently dropped: it wants its own row against the
smoke harness.

### Tests — `tests/openclaw_auth_layout_ladder_unit.sh`, 20/20 green

New arms:
- **8a** 2026.8.1 shape (old dir created-and-empty, credential + model pin in `openclaw.json`) → the
  create witness must NOT print the false issue, and must print its `ok:` line.
- **8b** pre-2026.8.1 shape → still `ok:` (fallback intact).
- **8c** 2026.8.1 with `auth.profiles` empty → the issue **still fires**. This is the arm that stops
  the fix from being "make the witness stop complaining". DIVE-3130.
- **9** `agent info`'s unpinned block resolves through the ladder. **Declared structural, not papered
  over**: grading it behaviourally means running `cmd_agent_info`, which reads systemd unit state, the
  on-disk registry and the enforced sudo grant — stubbing all three would grade the stubs. The
  resolver it now calls is graded behaviourally, on real fixtures, by arm 1.

### Mutation controls — every one asserted applied, then reverted

| mutation | failing assertions |
| --- | --- |
| **M6** un-wire the create witness (restore the hardcoded sqlite) | **2** — and 8a's failure message is verbatim the reported symptom: `issue:openclaw credential never reached the seat (…/openclaw-agent.sqlite is missing/empty)` |
| **M7** un-wire the `agent info` resolver | **1** |
| **M3** (quinn's) `if [[ "$type" == openclaw && -n "$_def" ]]` → `if false` | **2** |
| **M4** (quinn's) drop the openclaw clause in `agent_auth_health` | **1** |
| **M5** (quinn's) drop the openclaw clause in `auth_creds_present` | **2** |

Baseline 0 before and after every mutation. quinn's M3/M4/M5 were red on iteration 2 and are still red.

Thirteen neighbouring suites rc=0 on the patched tree: `agent_list_auth_health`,
`auth_status_per_agent_scope`, `auth_session_reaper`, `antigravity_auth_finalize`,
`selfcheck_cred_reached`, `selfcheck_unit`, `selfcheck_union`, `profile_seed_perms`,
`dive1762_regression`, `openclaw_seat_seed`, `openclaw_runtime`, `openclaw_pin_regrade`,
`type_map_registration_contract`.

### STILL OWED at merge — unchanged, and I am not signing it

The paid on-box 2026.8.1 acceptance has **not** run; I hold no box credential and the row's smoke box
has self-purged. `smoke-verified` **must not** be applied on the strength of the unit evidence above.
Rung 2's key path comes from this row's measurement of box 62.238.11.92, not from a re-measurement by
me. Named residual: the interactive device-login leg on 2026.8.1 was not measured, so whether it
writes `.auth.profiles` is unknown; if it does not, that seat resolves to no proven rung and the
unchanged mtime-poll path is what detects the login.

PR #772, opened 2026-09-05
fix(supervisor): mute the human leg for the no-output fleet-health class (DIVE-3982)

lodar:
## What

Extends DIVE-3970's human-leg mute to the **second** FLEET-HEALTH class, `no-output`
("N open row(s), nothing closed in Nd"). DIVE-3970 muted only `quota-exhausted`;
`no-output` still DMs lodar with a byte-identical human-facing template, and today's
(dev3, "9d") was a false positive — a healthy idle on-demand seat with an empty queue
(it woke and worked DIVE-3978 the moment work arrived).

## Fix

One clause added to `_sup_capacity_notify_human` (`src/cmd_supervisor.sh`): `no-output`
mutes the human leg **unconditionally**, ahead of the persistence escape.

- **Why unconditional (unlike quota):** `no-output` is not human-actionable — its remedy
  ("reassign or park") is main/ops triage, not lodar's — and its long duration is the
  benign idle case, not an incident the way a quota wall past its reset is. Persistence
  does **not** un-mute it (the deliberate difference from quota, which still escalates).
- **Machine leg unchanged:** `_sup_capacity_alert` stays unconditional, so main still
  triages every one and the DIVE-3272 cover is intact; if a seat is genuinely stuck,
  main escalates to lodar in plain language — exactly what was promised 2026-09-04.

## Tests (`tests/supervisor_escalate_delivery_unit.sh`)

Three existing assertions encoded the old no-output-pings-human behaviour (pure decision,
end-to-end tick, self-healing-kind case) — updated to muted. Added: no-output stays muted
even when `persisted=true`, and the machine leg still fires. Mutation-verified: removing
the guard line reds all four arms. Full harness green (PASS=127 FAIL=0).

## Release / residual

`5dive-cli` ships on release-cut, not on merge. This rides the **same next cut** as
DIVE-3970 (both merged to main, fleet on 0.25.10 predates both) so both classes quiet
lodar's phone together. Until that cut rolls both can still DM; dedup is
per-class-per-window and dev3 is now transacting, so imminent repeats are unlikely.

### DIFF
--- a/src/cmd_supervisor.sh
+++ b/src/cmd_supervisor.sh
@@ -1081,6 +1081,15 @@ _sup_capacity_alert() {  # <name> <class> <detail> [notify_human=true]
 # DIVE-3940's decision.
 _sup_capacity_notify_human() {  # <class> <quotaDeadline> [selfheal_kind] [persisted] -> true|false
   local class="${1:-}" qdl="${2:-}" kind="${3:-no}" persisted="${4:-false}"
+  # DIVE-3982: no-output mutes the human leg UNCONDITIONALLY — ahead of the
+  # persistence escape, unlike quota. Its remedy ("reassign or park") is
+  # main/ops triage, never lodar's, and a long duration is the benign idle case
+  # (an empty queue closes nothing), not an incident like a quota wall past its
+  # reset. The machine leg in _sup_capacity_alert stays unconditional, so main
+  # still triages every one and escalates to a human in plain language if one is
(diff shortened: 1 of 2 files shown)


lodar:
Duplicate of #771 (identical DIVE-3982 fix — 127/0, same guard, same position; only comment wording + one temp-var differ). #771 is merged; closing this so the row resolves.

PR #786, opened 2026-09-07
feat(plugin): dispatch a plugin's declared verb — contract §2's live half (DIVE-4035)

lodar:
Replaces #785, which I made unmergeable.

## What happened

#785 carried #779 (DIVE-4020) as a merge commit. I **squash**-merged #779, so its five commits are not ancestors of `main` — main has one equivalent-but-different commit instead. #785 therefore still carried them and went `CONFLICTING`. quinn's "land #779 and #785 shrinks to its own commit" assumed a merge, not a squash. My mistake, not the maker's.

## What this branch is

`87cff8dd` — the single DIVE-4035 commit quinn graded — cherry-picked onto current `main`. Nothing else.

## The check that makes it safe

The tree cannot be compared against the old branch (main has since gained #782 and #784, which that branch never had). So the invariant is on the **delta**, not the tree:

    git diff 87cff8dd^ 87cff8dd      ->  9 files, +776/-20
    git diff origin/main fix785      ->  9 files, +776/-20

Byte-identical except **4 lines of git blob index hashes**, all in `src/main.sh` — whose base moved because #782 touched it. Same textual change, applied cleanly to the newer base, no conflict.

So this branch introduces **exactly the delta quinn graded at `87cff8dd`**, and nothing more.

quinn: the grade transfers on that invariant, but the resolved tree is new — re-grade or confirm at your discretion.

### DIFF
--- a/changelog.d/DIVE-4035.md
+++ b/changelog.d/DIVE-4035.md
@@ -1,0 +1,42 @@
+## Unreleased — feat(plugin): a declared plugin verb is now dispatched (DIVE-4035)
+
+Contract v1 §2 let a plugin declare a `verb` capability and the enforcer validated the declaration
+— but nothing dispatched it. `plugins/voice` named the verb `voice`, `5dive plugin add voice@5dive`
+succeeded, no warning was printed, and `5dive voice` was still **unknown command**. One clause
+checked and not honoured, in the direction a publisher cannot see: §2's stated half ("an undeclared
+surface is inert") was enforced, its unstated half ("a DECLARED surface is live") was false.
+
+**What a plugin verb is invoked as** — the design call, now fixed and written into the contract. A
+verb resolves to an **executable file at a path 5dive computes**: `<plugin>/bin/<verb>`, `exec`'d
+with the caller's argv as an argument **vector**. The manifest supplies the NAME and nothing else.
+No command line, interpreter or entry-point path out of `plugin.json` ever reaches a shell — that is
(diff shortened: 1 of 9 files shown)


lodar:
**On-box end-to-end smoke of the headline, run against this branch's bundle at 975354ae — install as ROOT, dispatch as an UNPRIVILEGED user.**

That pairing is not decoration: [[an-atomic-mktemp-write-downgrades-the-file-to-0600]] records this exact feature shipping once with `plugin add` printing success while `5dive voice` answered `unknown command`, because every subverb was root-gated and there was no unprivileged reader yet. So the arm is the split, not the install.

Sandboxed with `STATE_DIR=<tmp>` so it touches nothing on the live box.

```
$ sudo STATE_DIR=$SB ./5dive plugin add voice
error: plugin add needs your confirmation and stdin is not a terminal — re-run with --yes

$ sudo STATE_DIR=$SB ./5dive plugin add voice --yes
OK — voice@5dive 1.0.0 installed — registers: channel verb

$ sudo -u agent-main STATE_DIR=$SB ./5dive voice --help
5dive voice — speech-to-text and text-to-speech on your own box
  …
rc=0
```

Modes on what the install wrote, since that is the regression: `installed.json` and `marketplaces.json` both **644**, not 600.

Also worth recording: the confirmation gate fires on a non-tty and refuses rather than proceeding, and the install prints the host-engine requirement as the publisher's own text rather than running it.

Harnesses on the merged tree: `plugin_verb_dispatch_unit.sh` 73/0, `plugin_contract_unit.sh` 95/0.

PR #787, opened 2026-09-07
DIVE-4064: un-blind the calibration harvester (the re-baseline is HELD — it reveals a real budget overrun)

lodar:
Two changes. The constant, and the one character class without which nobody can check the constant.

## The constant

`TIER_CAL_BASELINE_US` **119000 → 172120**.

`ubuntu-latest` drifted until probes read **151-153%** of 119000, past the **150% upper clamp**, so runs exit **6 UNDETERMINED** having never graded the corpus.

This is the **mirror** of the 173000 failure `tier.sh` already documents — that one sat too high, every probe read fast, and the lower clamp made the mechanism a silent no-op. A reference attached to an environment that moved fails at whichever clamp it drifts into; the low side is just louder, because it reds.

**It already blocked shipped work.** PR #786 (lodar-approved at the CODEOWNERS gate, quinn-graded PASS) cannot merge:

```
core-installed-host (1)  exit 6, 153%   192 harness rows, ZERO non-zero
test-installed-host      FAILED         <-- a REQUIRED context
```

`test-installed-host` is a pure aggregator (`[ "${d#*=}" = success ] || rc=1`), so an UNDETERMINED shard is laundered into a `failure` string while every harness in the run passed. **Two re-runs did not clear it** — that shard measured 153% twice, within 0.1%. It is not a lottery, it is reliably over the line.

## Why the harvester fix ships in the same PR

DIVE-2867 made this reference movable by a countable check, and justified that as *enforceable* because `tier-cal-harvest.sh` could feed `tier-cal-window.sh` from CI logs. **It could not.**

```
regex:  harness-budget\[[a-z]+/[a-z-]+\]     <- no digits
actual: harness-budget[core/installed-host-s2]: 192 harnesses, ...
```

Shard suffixes contain digits. Measured: **22 runs → 0 reports**, every one of them holding the lines. After the fix the same runs yield **4 reports each**.

And it fails **open**: an unparseable run reports `(no budget block)`, the same words as a run with genuinely no data. The tool's own closing warning points at workflow names, branch names and log retention — all external — so a correct warning still sends you the wrong way.

Without this fix the constant cannot be re-derived the way DIVE-2867 requires, by anyone. That is why it never had been.

## How 172120 was chosen

**Median of the 46 concordant readings** in a harvested **78-report window** (18 `unit-tests` pull_request runs, 2026-09-06..07).

Median of *all* concordant readings deliberately: the window is mildly bimodal (fast tail from 101305, working mode from ~169000) and splitting it needs an exclusion argued first — the estimator this file already refused: *"an outlier you have to argue away before you can compute is an outlier your estimator is too fragile to see."*

**I nearly got this wrong.** I had eight readings scraped off failing jobs pointing at ~178000. That is precisely the hand-copied handful DIVE-2867 exists to stop, and it is not what shipped.

## Checked

- At 172120 **all 46** concordant readings land inside the 100-150% clamp (worst 182924 → 106%; fast tail floors at 100%). **Zero would exit 6.**
- `tier.sh refadmit 172120 119000 <46 samples>` → `admit raise`. Note refadmit does **not choose** the value — raising is the one-sided safe direction so it admits *any* raise; the median is what selects it.
- `corpus_tier_budget_unit` 165/165 · `budget_attribution` 24/24 · `envelope_tier_provenance` 24/24 · `header_drift_window` 27/27 · `shellcheck -S error` clean.
- No test pins the literal constant (the `119000` hits are refadmit fixtures and fake-log fixtures), and the old non-sharded label still matches, so the harvester change is backward compatible.

## For the verifier

The window is reproducible: `scripts/tier-cal-harvest.sh <run-ids> --out=DIR` then `scripts/tier-cal-window.sh DIR/*.report`. Use explicit run ids from `workflows/unit-tests.yml/runs?event=pull_request` — `--branch=main` legitimately harvests nothing, because the sharded budget jobs only run on pull_request.

### DIFF
--- a/scripts/tier-cal-harvest.sh
+++ b/scripts/tier-cal-harvest.sh
@@ -121,10 +121,19 @@ parse_log() {
     blk="$(awk -F'\t' -v j="$job" '$1==j' "$log" | grep -a 'harness-budget\|runner; applied\|and .*% of the EFFECTIVE cap')"
 
     # "harness-budget[core/pristine]: 254 harnesses, 283s wall-clock, budget 300s (94% of budget)"
-    local totals; totals="$(printf '%s\n' "$blk" | grep -aoE 'harness-budget\[[a-z]+/[a-z-]+\]: [0-9]+ harnesses, [0-9]+s wall-clock, budget [0-9]+s \([0-9]+% of budget\)' | tail -1)"
+    #
+    # DIVE-4064: the label character class MUST admit digits. Real labels are SHARDED —
+    # "core/installed-host-s2", "core/pristine-s1" — and the original `[a-z-]+` cannot
+    # match them, so this harvested exactly NOTHING from every sharded run while
+    # reporting it as "(no budget block)", which is the same string a run with genuinely
+    # no budget data produces. Measured: 22 runs -> 0 reports, every one of them holding
+    # the lines; one character class later the same runs yield 4 reports each. That is
+    # why DIVE-2867's "the excuse is removed" was not true in practice and the reference
+    # had never been re-derived by this path.
+    local totals; totals="$(printf '%s\n' "$blk" | grep -aoE 'harness-budget\[[a-z]+/[a-z0-9-]+\]: [0-9]+ harnesses, [0-9]+s wall-clock, budget [0-9]+s \([0-9]+% of budget\)' | tail -1)"
     [[ -n "$totals" ]] || continue
     tier="$(printf '%s' "$totals"  | sed -E 's/^harness-budget\[([a-z]+)\/.*/\1/')"
-    label="$(printf '%s' "$totals" | sed -E 's/^harness-budget\[[a-z]+\/([a-z-]+)\].*/\1/')"
+    label="$(printf '%s' "$totals" | sed -E 's/^harness-budget\[[a-z]+\/([a-z0-9-]+)\].*/\1/')"
     harn="$(printf '%s' "$totals"  | sed -E 's/.*: ([0-9]+) harnesses.*/\1/')"
     wall="$(printf '%s' "$totals"  | sed -E 's/.*, ([0-9]+)s wall-clock.*/\1/')"
     budget="$(printf '%s' "$totals" | sed -E 's/.*budget ([0-9]+)s \(.*/\1/')"
--- a/tests/lib/tier.sh
+++ b/tests/lib/tier.sh
@@ -162,7 +162,44 @@ TIER_CAL_SCALE_MAX_PCT=150
 # from the box you happen to be on — that is exactly how 173000 got here, and the
 # failure mode is silent (every CI probe then reads fast, the floor clamps, and the
 # relative budget quietly stops existing while still printing a ratio).
-TIER_CAL_BASELINE_US=119000
+#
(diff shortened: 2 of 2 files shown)


lodar:
## The fix demonstrates itself on its own CI run

Live calibration from this PR's own `unit-tests` run (34154925689), which executes **with** the new baseline:

| shard | measured | @172120 (this PR) | @119000 (old) | verdict under old |
|---|---|---|---|---|
| core-installed-host (1) | 174315 | **101%** | 146% | ok |
| core-installed-host (2) | 181509 | **105%** | 152% | **EXIT 6 UNDETERMINED** |
| core-pristine (1) | 175438 | **101%** | 147% | ok |
| core-pristine (2) | 181018 | **105%** | 152% | **EXIT 6 UNDETERMINED** |

**Two of four shards on this very PR would have exited 6 under the old baseline.** All four sit at 101-105% against the new one — inside the 100-150% clamp with real headroom on both sides.

This is measurement on a live runner, not the projection from the harvested window. The window said the median concordant reading was 172120 and that zero of 46 would exceed the clamp; these four independent draws land where that predicted.

Status: **24 pass / 0 fail**, 8 still running (the `-confirm` and `full-*` shards).

Note the counterfactual column is arithmetic on the same measured microseconds, not a re-run — the runner draw is held fixed and only the divisor changes, which is exactly the claim being made.

lodar:
## Scope reduced: the re-baseline is BACKED OUT of this PR

I was wrong that this was shippable as one change. **This PR's own CI proved it.**

At 172120, `core-pristine-confirm (2)` graded cleanly — calibration 181072us/iter = **105%**, comfortably inside the clamp — and then failed on the corpus:

```
195 harnesses, ALL HARNESS-RC=0        <- nothing broke
OVER BUDGET by 3s (318s > 315s effective cap)
BUDGET CONFIRMATION — first sample 318s (106%), confirmed 318s (106%)
```

Reproduced twice at 318s, so not runner noise.

### What that means, and it is the real finding

Under the **old** baseline that same measurement is 181072/119000 = **152%** — past the 150% clamp, so the shard exited **6 UNDETERMINED** and no cap was ever applied. Raising the reference tightened the effective cap from *"up to 450s, or never graded"* down to **315s**. The corpus needs 318s.

**So the stale constant was masking a real budget overrun.** That is the instrument working as designed, and it matters more than the constant: this corpus does not fit its 300s budget on a normal runner, and nobody could see it.

But it means the re-baseline lands a **deterministic** red on a **required** context (`test-confirm`), which is worse for merge throughput than the intermittent exit-6 it replaces. So it cannot ship until the overrun is resolved.

### Not fixable by widening

Raising the budget, or the 150% clamp, to absorb 3s is DIVE-2525's ratchet wearing the calibration's clothes — and `tier.sh` refuses that direction on purpose. The choice is **trim the corpus** or **re-argue the budget with evidence**, and neither belongs inside a re-baselining change.

### What is left in this PR

The harvester fix only — the label class `[a-z-]+` becomes `[a-z0-9-]+` so sharded labels match. Pure bug fix, no behavioural surface, and it unblocks the only instrument that can derive the reference. Measured: 22 runs produced 0 reports before, 4 reports per run after.

The derivation stands and is reproducible for whoever picks the constant back up: median of 46 concordant readings in a 78-report window = **172120**, all 46 inside the clamp at that value.

PR #802, opened 2026-09-08
DIVE-4091 probe: the incident launcher under #801's arm must turn docker-install red (do not merge)

lodar:
Verifier probe for https://github.com/5dive-ai/5dive/pull/801. Same tree as #801's graded head bb204d5 with ONE change: 5dive-agent-start replaced by the v0.26.1 file that emptied every box on 2026-09-07 (the 'line 1687: dir: unbound variable' fatal). Expected: docker-install FAILS in the built image with 'FAIL: 5dive-agent-start exited 1' and a recorded tmux call list that contains kill-session but no new-session. That is the red direction of the arm measured in-image, which #801 could only show on a path-relocated copy. shellcheck going red here (SC2318) is expected and irrelevant. Close after reading.

### DIFF
--- a/.github/workflows/install-smoke.yml
+++ b/.github/workflows/install-smoke.yml
@@ -180,3 +180,85 @@ jobs:
             echo "ERROR: installed 5dive reports '$actual', expected version '$expected'"
             exit 1
           fi
+      # DIVE-4091: nothing in CI ever STARTED an agent. v0.26.1 emptied every box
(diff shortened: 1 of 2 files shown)


lodar:
Read. docker-install failed in 58s with the incident's fatal ('line 1687: dir: unbound variable'), tmux calls recorded = kill-session only, arm printed 'FAIL: 5dive-agent-start exited 1'. That is the red direction of #801's arm measured in the built image. Receipt recorded on DIVE-4091. Closing; not for merging.

PR #887, opened 2026-09-11
test(heartbeat): bound the wake-failure scan to its branch and drop comments (DIVE-4299)

lodar:
## What

`tests/agent_list_overdue_liveness_unit.sh` grades a real invariant — **an undeliverable wake must not stamp `_hb_mark_seen`**, because stamping a failed wake makes the OVERDUE alarm unreachable. It graded it with a raw source-text scan:

```
awk '/wake failed — will retry next tick/{found=1} found && /_hb_mark_seen/{c++} END{print c+0}' src/cmd_heartbeat.sh
```

That trips `found` on the **first line containing the sentence — a comment reads identically to code** — and then counts to **end of file**.

## Why it matters (measured, not felt)

DIVE-4279's own header comment quoted the incident log verbatim ~4400 lines above the wake function. Count went `0 -> 3`, re-grading three untouched calls. **The merge queue ejected PR #874 twice** (added 08:24:39Z, removed 08:37:48Z, re-added 08:40:57Z), each enqueue burning a candidate CI run, and it cost DIVE-4279 an extra iteration — for an invariant that diff never touches (`git diff 6ec89c79...32373ed4 -- src/cmd_heartbeat.sh | grep -c '^+.*_hb_mark_seen'` = **0**).

#874 reflowed its comment off the literal, which fixed that branch. The wire stayed live for the next person to quote the sentence. Same class as DIVE-3591 / DIVE-4284.

## The fix

Comment-stripping alone would have stopped #874 but not the next shape — **the EOF window is a second, independent defect**. Both go:

- **(a)** whole-line comments are dropped — the invariant is about CODE, not prose;
- **(b)** the window is bounded to the wake-failure **branch** by indentation, walking **both** directions from the anchor. Forward-only is half an arm: a stamp placed *above* the log inside the same `else` is the likeliest real violation and reads as 0;
- **(c)** a branch collapsed onto its own `if`/`else` line is graded **alone** — it carries the enclosing block's indent, so expanding it swallows the siblings and re-creates the false positive in the small (found by self-audit after the first commit: a clean collapsed branch read `1`);
- **(d)** a missing anchor prints `NO-ANCHOR`, not a passing `0` — renaming the log message is exactly the edit someone touching this path makes, and the old scan greened on it.

## Also fixed: the sibling grep two lines above it

`n_seen=$(grep -c '_hb_mark_seen "$name"' ...)` is the same class in the same block — a comment quoting the call inflates it exactly as a call does, on the same file, and it is the one that moved 3 -> 4 in control B below. Left live it is the next false ejection. Comments dropped, pinned with an arm that prepends such a comment to the real source and requires the count to stay 3.

## How it was checked

**Suite:** `agent_list_overdue_liveness_unit.sh` **33/0 -> 41/0** (8 new arms).

**Scanner pinned on fixtures, both directions** — an arm that cannot fail grades nothing:

| fixture | expect |
|---|---|
| comment quotes the sentence, code stamps nothing (the #874 shape) | `0` |
| stamp on an unrelated path after the branch closes (the EOF bug) | `0` |
| stamp inside the branch, **above** the log line | `1` |
| stamp inside the branch, **below** the log line | `1` |
| log line renamed away | `NO-ANCHOR` |
| branch collapsed onto one line, clean, unrelated stamp after | `0` |
| branch collapsed onto one line, violation **on** that line | `1` |

**Controls on the real tree, not just fixtures:**

- **A — reproduce the ejection.** Re-injected the DIVE-4279 comment at line 1729 of `src/cmd_heartbeat.sh`: **old scan reads 3** (the exact red that ejected #874); **new arm reads 0**, suite 41/0. False positive dead.
- **B — the arm still grades.** Injected a genuine `_hb_mark_seen` into the wake-failure branch: new arm **FAILS, want 0 got 1** (and the adjacent stamp-count arm correctly moved 3 -> 4). Restored; tree clean.

**Regression:** `heartbeat_send_line_reason_unit` 28/0 · `heartbeat_stall_sweep_unit` 67/0.

**Pre-push rail:** green (title / shellcheck / 1-of-1 changed harness).

**shellcheck:** `-S error` **clean (0)** and `--include=SC2318` **clean (0)** — the two gated passes. The 5 added findings are all `SC2016` (info, "expressions don't expand in single quotes") on the awk program and fixture heredocs — intentional, same class already present on base, and below both gated passes.

**Scope:** test-file only. `src/cmd_heartbeat.sh` is untouched (`git diff --stat` below). Not a provisioning-path change — no `scripts/inc/5dive-cli.sh`, no agent-create path, no `sweepKeySync`.

## What stays uncovered (signing this)

The arm is still a **source-shape** scan, not a behavioural one. A stamp reached *indirectly* from the wake-failure branch — a helper called there that stamps internally — is invisible to it, exactly as it was before this PR. Closing that needs the branch extracted into a callable function and graded by running it, which changes `cmd_heartbeat.sh` on a hot path and is not what this row asked for. Named here rather than left implied.

## Note for reviewers

The fixtures put the guarded sentence **into the test file itself**. That is safe only because the scanner takes its target as an argument and is pointed at `src/cmd_heartbeat.sh`; a guard that scans a fixed path or the whole repo could not carry its own fixtures. Repo-wide grep confirms the only other mention is a pre-existing comment in `tests/heartbeat_send_line_reason_unit.sh:11`, which nothing scans.

Knowledge compiled: `community/wiki/a-comment-that-names-a-guarded-literal-enters-the-guards-match-set.md` (new dated section + index line updated — the page previously said this wire was still live).

Closes DIVE-4299.


lodar:
**Iteration 2 — merged `origin/main` (406bfb2f) and LAYERED the two half-fixes.** Head `7d391874`.

The bounce was right and the reconciliation is not "pick a side": this arm had **three independent defects** and each row had fixed a strict subset.

| | anchor | window | comment-blind count |
|---|---|---|---|
| DIVE-4310 (on main) | fixed | still EOF | still counts comments |
| DIVE-4299 (prev head) | **still prose** | fixed | fixed |
| this head | fixed | fixed | fixed |

Measured, all on the merged tree unless stated:

- **suite 53/0** (main alone 44, previous head 41). No `NO-ANCHOR` anywhere in the output except as the *name* of the arm that asserts it, which passes.
- Main's two newer arms kept verbatim — *the wake-failure verdict is exactly one code line*, *and the verdict names the step that failed* (`_HB_WAKE_FAIL_REASON`).
- **Control A** (a comment quoting the retired log sentence **and** a comment naming `_hb_mark_seen`, both injected below the anchor in `src/cmd_heartbeat.sh`, restored byte-identical after): this head **53/0**; the original DIVE-4278 scan reads **4**; **main's current arm reads 1** — the wire is live on main today, as the bounce said.
- **Control B** (a genuine `_hb_mark_seen` inside the wake-failure branch): **50/3** with the stamp *above* the counter and **50/3** *below* it. The arm still grades in both directions.
- Regressions: `heartbeat_send_line_reason_unit` 28/0, `heartbeat_stall_sweep_unit` 67/0. `shellcheck -S error` rc=0, `shellcheck --include=SC2318` rc=0.

**The one design decision worth arguing with.** The anchor is loosened from `^[[:space:]]*sk_fail=…` to a statement boundary `(^|[[:space:];&|])sk_fail=…`, because the strict prefix reads `NO-ANCHOR` on a healthy file whose branch is collapsed onto one `if … else … fi` line. That moves the comment-proofness from the `^` prefix onto the awk **comment map** (`!cmt[NR]`) — the two guards are not interchangeable and dropping either re-opens a different hole. Main's strict `grep -c … == 1` is **kept as its own arm**: it pins today's shape, the tolerant scanner survives a reflow, neither replaces the other.

**Pushed on an audited five-clause `FIVE_PUSH_OVERRIDE`, not `--no-verify`.** The pre-push rail grades the whole merge range (31 harnesses) and two of main's are red — `task_merge_gate_result_pr_unit` 33/1, `task_merge_gate_selftest_unit` 12/2. Both reproduce with **identical counts and identical arm names** on a clean detached worktree at `406bfb2f` with no part of this branch in it. Zero head-only failures.

**Residual, unchanged and signed again:** still a source-SHAPE scan, so a stamp reached indirectly via a helper called from the branch is invisible — strictly no worse than the pre-fix state. Closing it means extracting the branch into a callable function and grading it behaviourally, which edits `cmd_heartbeat.sh` on a hot path.

lodar:
**Set-diff proof that no side was dropped** (stronger than the arm count, which only says it grew): comparing arm NAMES, `origin/main` 406bfb2f has 44 and this head has 53, and `comm -23 main head` is **empty** — every one of main's arms survives verbatim, 9 are new, 0 dropped.

